PR #16, opened 2020-02-21
Feature/seed scripts

geekbeast:
Put together some scripts to make it easier to get a stack with data in it.


scripts/ol_init.sh
```diff
@@ -0,0 +1,3 @@
+curl -XPUT http://localhost:9200/_cluster/settings -H 'Content-type: application/json' --data-binary $'{"transient":{"cluster.max_shards_per_node":10000}}'
```
katherinebernstein: It might be useful to add an ol_wipe script that deletes all your indices + drops/recreates your database? I can imagine people calling init expecting that it would reset your setup if there's no separate wipe script

jokedurnez:
I added a PR into this branch with fixes.

PR #30, opened 2020-08-10
Feature/more scripts

geekbeast:
Most of these scripts changes are only use for people running Ubuntu.

- Automatically generate and run sql commands for dropping all roles (useful for cleaning up atlas).
- Automatically generate and run sql commands for truncating all tables (useful for not dropping db when citus is not in docker)

Some of my local drop scripts got captured, but they'll be overwritten when you run primary scripts.


geekbeast:
Merging for release.

PR #44, opened 2020-11-18
get the external database table / atlas data set schema

UnsungHero97:


conductor-client/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -244,6 +244,23 @@ class ExternalDatabaseManagementService(
...
+
+        val table = getOrganizationExternalDatabaseTable(tableId)
+        val sql = getExternalDatabaseTableSchemaSql(table.name)
+        val schemas = BasePostgresIterable(
+                StatementHolderSupplier(externalDbManager.connectToOrg(organizationId), sql)
+        ) { rs ->
+            rs.getString(PostgresColumn.PG_SCHEMA_NAME)
+        }.toList()
+
+        if (schemas.size != 1) {
```
katherinebernstein, line 257: this seems unnecessary, unless you specifically want to return null in the case that there are multiple results

conductor-client/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -731,6 +748,11 @@ class ExternalDatabaseManagementService(
         return "SELECT * FROM $tableName"
     }
 
+    private fun getExternalDatabaseTableSchemaSql(tableName: String): String {
+        return "SELECT ${PostgresColumn.PG_SCHEMA_NAME} FROM pg_catalog.pg_tables " +
+                "WHERE ${PostgresColumn.PG_TABLE_NAME} = '${tableName}'"
```
katherinebernstein, line 753: this should be done as a PreparedStatement otherwise we're opening ourselves up to SQL injection

conductor-client/src/main/java/com/openlattice/postgres/PostgresColumn.java
```diff
@@ -400,6 +400,8 @@
     public static final String                   PERMISSIONS_FIELD                 = "permissions";
     public static final PostgresColumnDefinition PERMISSIONS                       =
             new PostgresColumnDefinition( PERMISSIONS_FIELD, TEXT_ARRAY );
+    public static final String                   PG_SCHEMA_NAME                    = "schemaname";
+    public static final String                   PG_TABLE_NAME                     = "tablename";
```
makosblade, line 404: Prefer `PG_SCHEMA_FIELD` as with the other column labels

makosblade: And same for `PG_TABLE_FIELD`

katherinebernstein, line 404: not at all important, but these feel like they should live somewhere separate since they're postgres internal values and are not being used to construct a PostgresColumnDefinition

makosblade (dismissed):
Couple of small comments, otherwise looks good to me

conductor-client/src/main/java/com/openlattice/postgres/PostgresColumn.java
makosblade: Oh yeah this may be better than my comment

conductor-client/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
UnsungHero97: That's the intention. I don't know how to handle that edge case, so returning null seemed the most reasonable.

conductor-client/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
UnsungHero97: Can you point me to an example? Should the other functions like this also be changed?

conductor-client/src/main/java/com/openlattice/postgres/PostgresColumn.java
UnsungHero97: I agree, just didn't know where to put them. I can create a separate class/

makosblade: Yeah probably a PostgresInternalColumns.kt?

conductor-client/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
makosblade: You'd convert the `{tableName}` to `?` and then use a `PreparedStatementHolder` instead of `StatementHolder`

makosblade (dismissed):

conductor-client/src/main/java/com/openlattice/postgres/PostgresColumnsInternal.java
```diff
@@ -0,0 +1,10 @@
+package com.openlattice.postgres;
+
+public class PostgresColumnsInternal {
```
katherinebernstein, line 3: all new classes should be written in kotlin

katherinebernstein: for reference:
```
package com.openlattice.postgres

class PostgresColumnsInternal {

    companion object {
        const val SCHEMA_NAME = "schemaname"
        const val TABLE_NAME = "tablename"
    }

}
```

katherinebernstein (approved):

makosblade (approved):

PR #45, opened 2020-11-18
close resultsets

makosblade:
gotta close em


conductor-client/src/main/java/com/openlattice/edm/properties/PostgresTypeManager.java
```diff
@@ -125,13 +125,13 @@ public PostgresTypeManager( HikariDataSource hds ) {
...
             List<PropertyType> result = Lists.newArrayList();
-            ResultSet rs = ps.executeQuery();
-            while ( rs.next() ) {
-                result.add( ResultSetAdapters.propertyType( rs ) );
+            try ( ResultSet rs = ps.executeQuery(); ) {
+                while( rs.next() ) {
+                    result.add( ResultSetAdapters.propertyType( rs ) );
+                }
+                connection.close();
+                return result;
```
UnsungHero97, line 133: All of these all look nearly identical. Is it possible to extract a common function to handle the try + while block of logic? In JS, we can do something like this (pseudocode):

```
function close(ps, callback) {
  try {
    ResultSet rs = ps.executeQuery();
    while( rs.next() ) {
      callback(rs);
    }
    connection.close();
  }
  catch () { ... }
  finally { ... } 
}

function getPropertyTypes() {
  ...
  PreparedStatement ps = ...
  List<PropertyType> result = ...
  close(ps, (rs) => {
    result.add( ResultSetAdapters.propertyType( rs ) );
  });
  return result;
}

```

makosblade: Yeah, can definitely do that. I was just trying to spend as little time on this code as possible right now because its quite old and should likely be replaced with Kotlin updated code

katherinebernstein (approved):

conductor-client/src/main/java/com/openlattice/edm/properties/PostgresTypeManager.java
katherinebernstein: ( + should someday be replaced by hazelcast-only calls instead of talking to postgres directly)

UnsungHero97 (approved):

conductor-client/src/main/kotlin/com/openlattice/edm/properties/PostgresTypeManager.kt
```diff
@@ -0,0 +1,133 @@
...
+                        while (rs.next()) {
+                            result.add(ResultSetAdapters.id(rs))
+                        }
+                        connection.close()
+                        return result
+                    }
+                }
+            }
+        } catch (e: SQLException) {
+            logger.debug("Unable to load ids for query: {}", query, e)
```
katherinebernstein, line 74: should probably be logger.error

conductor-client/src/main/kotlin/com/openlattice/edm/properties/PostgresTypeManager.kt
```diff
@@ -0,0 +1,133 @@
...
+                Predicates.equal<UUID, AssociationType>(AssociationTypeMapstore.SRC_INDEX, entityTypeId),
+                Predicates.equal<UUID, AssociationType>(AssociationTypeMapstore.DST_INDEX, entityTypeId)
+        )).stream()
+    }
+
+    fun getAssociationTypeIds(): Iterable<UUID> {
+        return associationTypes.keys
+    }
+
+    fun getEnumTypeIds(): Stream<UUID>? {
```
katherinebernstein, line 102: is this supposed to be nullable?

conductor-client/src/main/kotlin/com/openlattice/edm/properties/PostgresTypeManager.kt
```diff
@@ -0,0 +1,133 @@
...
+            return ImmutableList.of()
+        }
+    }
+
+    fun getEntityTypes(): Iterable<EntityType> {
+        return entityTypes.values
+    }
+
+    fun getEntityTypesStrict(): Iterable<EntityType> {
+        return entityTypes.values(Predicates.equal<UUID, EntityType>(EntityTypeMapstore.CATEGORY_INDEX, SecurableObjectType.EntityType.name))
```
katherinebernstein, line 84: I don't think you need the `.name` here -- it should match on the enum value itself (not sure if .name will work actually)

conductor-client/src/main/kotlin/com/openlattice/edm/properties/PostgresTypeManager.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    fun getAssociationEntityTypes(): Iterable<EntityType> {
+        return entityTypes.values(Predicates.equal<UUID, EntityType>(EntityTypeMapstore.CATEGORY_INDEX, SecurableObjectType.AssociationType.name))
```
katherinebernstein, line 88: (same thing as above)

conductor-client/src/main/kotlin/com/openlattice/edm/properties/PostgresTypeManager.kt
```diff
@@ -0,0 +1,133 @@
...
+        while (!idsToLoad.isEmpty()) {
+            val id = idsToLoad.poll()
+            getEntityTypeChildrenIds(id).forEach(Consumer { e: UUID -> idsToLoad.add(e) })
+            children.add(id)
+        }
+        return StreamUtil.stream(children)
+    }
+
+    override fun getAllPropertyTypesInSchema(schemaName: FullQualifiedName): Set<UUID> {
+        return propertyTypes.keySet( Predicates.equal<UUID, PropertyType>(PropertyTypeMapstore.SCHEMAS_INDEX, schemaName.fullQualifiedNameAsString))
```
katherinebernstein, line 123: (same thing as above re converting to string)

UnsungHero97 (approved):

katherinebernstein (approved):

PR #46, opened 2020-11-19
reconstruct closed prs

katherinebernstein:


conductor-client/src/main/kotlin/com/openlattice/hazelcast/serializers/AclKeyStreamSerializer.kt
```diff
@@ -0,0 +1,46 @@
...
+    companion object {
+
+        @JvmStatic
+        fun serialize(out: ObjectDataOutput, `object`: AclKey) {
+            ListStreamSerializers.fastUUIDListSerialize(out, `object`)
+        }
+
+        @JvmStatic
+        fun deserialize(`in`: ObjectDataInput): AclKey {
+            return AclKey(*ListStreamSerializers.fastUUIDArrayDeserialize(`in`))
```
makosblade, line 23: * does an extra array creation when this AclKey() constructor is already defensive and will do its own copy. So would be better to use the `fastUUIDListDeserialize`

makosblade (dismissed):

UnsungHero97 (dismissed):

makosblade (approved):

UnsungHero97 (approved):

PR #58, opened 2020-12-04
move logic to initialize org db out of entry processor to avoid cross…

katherinebernstein:
…-partition hazelcast calls


conductor-client/src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -250,8 +250,23 @@ class Assembler(
 
     fun createOrganization(organizationId: UUID, dbName: String) {
         createOrganizationTimer.time().use {
-            assemblies.set(organizationId, OrganizationAssembly(organizationId))
-            assemblies.executeOnKey(organizationId, InitializeOrganizationAssemblyProcessor(dbName).init(acm))
+
+            if (!assemblies.tryLock(organizationId)) {
```
makosblade, line 254: Pretty sure this will NPE b/c the key won't be in this map yet. May be better to move a portion of the logic in `acm.createAndInitializeOrganizationDatabase` into here or just create an entry in the organizations map and then do this try lock on that map

katherinebernstein: ooh true

makosblade (approved):

UnsungHero97 (approved):

PR #64, opened 2020-12-12
Update GeocoderTransform.java

kengie:
Small change to url

### DIFF
--- a/conductor-client/src/main/java/transforms/GeocoderTransform.java
+++ b/conductor-client/src/main/java/transforms/GeocoderTransform.java
@@ -19,7 +19,7 @@ import java.util.Map;
 import java.util.Optional;
 
 public class GeocoderTransform extends Transformation<Object> {
-    protected static final String       NOMINATIM_SERVICE_URL = "https://osm.openlattice.com";
+    protected static final String       NOMINATIM_SERVICE_URL = "https://osm.openlattice.com/";
     private final          String       addressObject;
     private final          GeocodingApi geocodingApi;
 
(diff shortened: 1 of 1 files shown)


UnsungHero97 (approved):

makosblade (approved):

conductor-client/src/main/java/transforms/GeocoderTransform.java
```diff
@@ -19,7 +19,7 @@
 import java.util.Optional;
 
 public class GeocoderTransform extends Transformation<Object> {
-    protected static final String       NOMINATIM_SERVICE_URL = "https://osm.openlattice.com/nominatim/";
+    protected static final String       NOMINATIM_SERVICE_URL = "https://osm.openlattice.com";
```
makosblade, line 22: we should probably move this into a config file some time in the future

nathanlmeyers: I think we need it to be: `https://osm.openlattice.com/` so the requests end up like:
`https://osm.openlattice.com/search?q=94118` (plus the other arguments)

Everything was previously of the form `https://osm.openlattice.com/nominatim/search?q=94118`

nathanlmeyers (changes requested):

conductor-client/src/main/java/transforms/GeocoderTransform.java
kengie: sounds good, I was wondering if we needed the slash but wasn't sure.

UnsungHero97 (approved):

makosblade (approved):

PR #77, opened 2021-01-05
Fix/lattice 2752 update search all users

solomontang:
there definitely seems like there's a better way to provide a default value for the `search_engine` param. I could not find any documentation to do this with retrofit.


conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
```diff
@@ -47,7 +47,7 @@ class Auth0UserDirectoryService(
 
         var page = 0
         var pageOfUsers = auth0ManagementApi.searchAllUsers(searchQuery, page++,
-                                                            DEFAULT_PAGE_SIZE
+                                                            DEFAULT_PAGE_SIZE, "v3"
```
katherinebernstein, line 50: It might be good to store like `AUTH0_API_VERSION` as a static constant in Auth0ManagementApi that's referenced here just so that piece is tightly coupled and to reduce the number of hardcoded strings

UnsungHero97 (approved):

anzioka (approved):

PR #82, opened 2021-01-11
Converts the organization controller to Kotlin.

geekbeast:
Behavior of API renames unchanged to maintain backwards compatibility with one exception-- getOrganizationEntitySets now returns an error instead of null when organization is not found.

Ignore JdbcConnection and DatasourcesApi for now.


api/src/main/kotlin/com/openlattice/organizations/DatasourcesApi.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ *
+ * @author Matthew Tamayo-Rios &lt;matthew@openlattice.com&gt;
+ */
+interface DatasourcesApi {
+    companion object {
+        // @formatter:off
+        const val SERVICE = "/datastore"
+        const val CONTROLLER = "/organization"
```
UnsungHero97, line 15: We have `/datastore/organizations` and with this `/datastore/organization`. Any reason why these have to be named so similarly? Why isn't the controller `/datasources`?

api/src/main/kotlin/com/openlattice/organizations/DatasourcesApi.kt
```diff
@@ -0,0 +1,44 @@
...
+        const val BASE = SERVICE + CONTROLLER
+        // @formatter:on
+
+        const val ID = "id"
+        const val ID_PATH = "/{$ID}"
```
UnsungHero97, line 20: I think it'd be clearer if these are `ORG_ID`.

api/src/main/kotlin/com/openlattice/organizations/JdbcConnection.kt
```diff
@@ -0,0 +1,23 @@
+package com.openlattice.organizations
+
+import com.fasterxml.jackson.annotation.JsonFilter
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.openlattice.client.serialization.SerializationConstants
+import java.util.*
```
UnsungHero97, line 7: Does this have to be `.*`?

api/src/main/kotlin/com/openlattice/organizations/DatasourcesApi.kt
geekbeast: Ignore this.

api/src/main/kotlin/com/openlattice/organizations/JdbcConnection.kt
geekbeast: Ignore this.

api/src/main/kotlin/com/openlattice/organizations/DatasourcesApi.kt
geekbeast: It's ID in organization api and these will be merged into that. Ignore this.

datastore/src/main/kotlin/com/openlattice/organizations/controllers/OrganizationsController.kt
```diff
@@ -0,0 +1,902 @@
+package com.openlattice.datastore.services.com.openlattice.organizations.controllers
```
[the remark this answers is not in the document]
geekbeast: Weird. Also weird that it compiled.

UnsungHero97 (approved):

makosblade (approved):

PR #89, opened 2021-01-15
Feature/store table schema

katherinebernstein:


makosblade (approved):

conductor-client/src/main/kotlin/com/openlattice/organizations/mapstores/OrganizationExternalDatabaseTableMapstore.kt
```diff
@@ -15,6 +15,9 @@ import java.sql.PreparedStatement
 import java.sql.ResultSet
 import java.util.*
 
+const val NAME_INDEX = "name"
+const val SCHEMA_INDEX = "schema"
```
UnsungHero97, line 19: 1. These don't seem to be used in this class. Are they used somewhere else? If so, is this the best place to hold these?
2. Is there a reason you can't use the `NAME` and `SCHEMA` constants from `SerializationConstants`?

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
```diff
@@ -95,138 +99,73 @@ class BackgroundExternalDatabaseSyncingService(
...
+                Optional.empty(),
+                orgId,
+                oid,
+                schemaName
+        )
 
-            edms.deleteOrganizationExternalDatabaseColumnObjects(missingColumnsByTable)
-            totalSynced += missingColumnIds.size
+        if (reservationService.isReserved(table.getUniqueName())) {
+            return organizationExternalDatabaseTables.getValue(reservationService.getId(table.getUniqueName()))
```
UnsungHero97, line 152: No change required, just a thought. You could use a variable to store the value of `table.getUniqueName()` and avoid invoking it twice.

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
```diff
@@ -95,138 +99,73 @@ class BackgroundExternalDatabaseSyncingService(
...
         }
 
-        return totalSynced
+        createSecurableTableObject(orgOwnerAclKeys, orgId, table)
+
+        return table
```
UnsungHero97, line 157: `createSecurableTableObject` returns `newTableId`. Don't you need to update `table` with the new id prior to returning it?

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
```diff
@@ -238,48 +177,137 @@ class BackgroundExternalDatabaseSyncingService(
...
+        val storedColumns = organizationExternalDatabaseColumns.getAll(existingColumnIdsByName.values.toSet())
+
+        val fullExistingColumns = mutableListOf<OrganizationExternalDatabaseColumn>()
+
+        val updatedColumns = storedColumns.mapNotNull { (id, storedColumn) ->
+            val currentColumn = currentColumnsById.getValue(id)
+            val updatedColumn = OrganizationExternalDatabaseColumn(
+                    id = storedColumn.id,
+                    name = currentColumn.name,
+                    title = storedColumn.title,
```
UnsungHero97, line 248: What's the logic for using `currentColumn` vs `storedColumn`? `name` should be immutable and `title` is mutable, right? In which case, shouldn't it be `name = storedColumn.name` and `title = currentColumn.title`?

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
```diff
@@ -238,48 +177,137 @@ class BackgroundExternalDatabaseSyncingService(
...
+            )
+
+            fullExistingColumns.add(updatedColumn)
+
+            if (updatedColumn == storedColumn) null else updatedColumn
+        }.associateBy { it.id }
+
+        if (updatedColumns.isNotEmpty()) {
+            organizationExternalDatabaseColumns.putAll(updatedColumns)
+        }
```
UnsungHero97, line 264: If `updatedColumns` is empty, would that mean that all the columns were removed? In which case, shouldn't hat also be reflected in the update?

conductor-client/src/main/kotlin/com/openlattice/hazelcast/processors/organizations/ExternalDatabaseTableEntryProcessor.kt
```diff
@@ -0,0 +1,34 @@
...
+        return null
+    }
+
+    data class Result(val value: Any? = null, val modified: Boolean = true)
+}
+
+
+
+
+
```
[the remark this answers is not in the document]
katherinebernstein: oh weird, removed

conductor-client/src/main/kotlin/com/openlattice/hazelcast/serializers/ExternalDatabaseTableEntryProcessorStreamSerializer.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override fun write(out: ObjectDataOutput, obj: ExternalDatabaseTableEntryProcessor) {
+        val baos = ByteArrayOutputStream()
+        val oos = ObjectOutputStream(baos)
+        oos.writeObject(obj)
+        oos.close()
+        oos.flush()
+        baos.close()
+
+        out.writeByteArray(baos.toByteArray())
```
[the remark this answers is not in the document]
katherinebernstein: ehh if this fails then the serialization would explode either way. This is also just following the same pattern we implement elsewhere

conductor-client/src/main/kotlin/com/openlattice/organizations/mapstores/OrganizationExternalDatabaseTableMapstore.kt
katherinebernstein: ooh yeah these should be part of the map index config -- they're referenced elsewhere but adding that index will improve perf there.

SerializationConstants is used for Jackson serialization, and these are for hazelcast map indices which are basically shortcuts to the getters, so assuming those will be identical seems dangerous.

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
katherinebernstein: updated

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
katherinebernstein: the id doesn't actually change

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
katherinebernstein: no, the goal here is to not overwrite any metadata users have added -- if they add a description for a column through the UI, and that column is synced, we don't want that description to disappear.

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
katherinebernstein: no, the cleanup not done in this function

UnsungHero97 (dismissed):

UnsungHero97 (approved):

makosblade (approved):

PR #93, opened 2021-01-20
columns should be plural

UnsungHero97:


geekbeast (approved):
This should be fine, since this only would be an issue on entity set creation.

geekbeast:
This is not a merge into master or develop.

PR #95, opened 2021-01-21
include load filtered entity set + integration.openlattice.com

yue-ol:
updating openlattice.yaml to include 
* integration.openlattice.com
* new endpoint for load filtered entity set
* + fix for entity set object for new flags

### DIFF
--- a/api/openlattice.yaml
+++ b/api/openlattice.yaml
@@ -2445,6 +2445,44 @@ paths:
       responses:
         "200":
           description: "Success"
+  /datastore/data/set/{entitySetId}/filtered:
+    post:
+      summary: Loads data in multiple pages
+      operationId: loadFilteredEntitySetData
+      parameters:
+        - name: entitySetId
+          in: path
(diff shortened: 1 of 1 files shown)


api/openlattice.yaml
```diff
@@ -2503,6 +2505,45 @@ paths:
...
+          required: true
+          schema:
+            type: string
+            format: uuid
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: array
```
katherinebernstein, line 2524: Just want to double check that "array" is expected here -- the body should just be a single FilteredDataPageDefinition object

geekbeast: Also think this should be object.

geekbeast (changes requested):
Double check where there request body for FilteredDataPage definition should be an array.

geekbeast (approved):

katherinebernstein (approved):

PR #100, opened 2021-01-25
filter out aces with empty permissions

UnsungHero97:
Aces with empty permissions are effectively the same as not having any permissions, but the ace object still exists and the API still returns these objects. It seems unnecessary for the API to return aces with empty permissions.

As far as I can tell, the only callers of `getAllSecurableObjectPermissions` are `PermissionsController`, which is what this change is intended for, and `AuditRecordEntitySetsManager`, which I'm not sure if this change will have an adverse impact on:

https://github.com/openlattice/openlattice/blob/develop/conductor-client/src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt#L206

https://github.com/openlattice/openlattice/blob/develop/conductor-client/src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt#L327

https://github.com/openlattice/openlattice/blob/develop/conductor-client/src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt#L331

https://github.com/openlattice/openlattice/blob/develop/conductor-client/src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt#L337

I'm also not entirely sure if this is the correct level in which to be making this change. Perhaps I should do this in `PermissionsController`, or add a boolean query parameter to the request to optionally do the filter. Thoughts?

### DIFF
--- a/conductor-client/src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
+++ b/conductor-client/src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
@@ -417,7 +417,7 @@ class HazelcastAuthorizationService(
     @Timed
     override fun getAllSecurableObjectPermissions(key: AclKey): Acl {
         val acesWithPermissions = aces.entrySet(hasAclKey(key))
-                .filterNot { it.value.isEmpty() }
+                .filter { it.value.isNotEmpty() }
                 .map { Ace(it.key.principal, it.value.permissions) }
                 .toSet()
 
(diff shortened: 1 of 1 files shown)


conductor-client/src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
```diff
@@ -426,6 +427,7 @@ class HazelcastAuthorizationService(
     @Timed
     override fun getAllSecurableObjectPermissions(keys: Set<AclKey>): Set<Acl> {
         return aces.entrySet(hasAnyAclKeys(keys))
+                .filterNot { it.value.isEmpty() }
```
katherinebernstein, line 430: the most minor of nits, but I think
```
.filter { it.value.isNotEmpty() }
```
feels more natural and is a bit easier to read than
```
.filterNot { it.value.isEmpty() }
```

katherinebernstein (approved):

geekbeast (approved):

makosblade (approved):

conductor-client/src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
UnsungHero97: I agree, that's much better; didn't know `isNotEmpty()` was available.

katherinebernstein (approved):

makosblade (approved):

PR #104, opened 2021-01-26
actually speed up neighbors pre-filtering

katherinebernstein:


makosblade (approved):

geekbeast (approved):

conductor-client/src/main/kotlin/com/openlattice/datastore/services/EntitySetManager.kt
```diff
@@ -108,4 +109,10 @@ interface EntitySetManager {
 
     fun setupOrganizationMetadataAndAuditEntitySets(entitySet: EntitySet)
 
+    fun getAuthorizedNeighborEntitySetsToEntityTypes(
```
UnsungHero97, line 112: I'm struggling to understand this name. What does `ToEntityTypes` mean? Using `To` usually implies a map, like A maps to B, but this is not returning a map. I recommend renaming.

UnsungHero97 (approved):

conductor-client/src/main/kotlin/com/openlattice/datastore/services/EntitySetManager.kt
katherinebernstein: Ooh good call, I originally had this returning a map but that was since updated. I'll change the name

makosblade (approved):

UnsungHero97 (approved):

PR #88, opened 2021-01-15
Feature/pluggable warehouse support

geekbeast:
Work in progress to support cataloging and managing permissions in external data warehouses.

PR #115, opened 2021-02-13
updates to metadata entity sets

UnsungHero97:


conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -108,10 +98,33 @@ class OrganizationMetadataEntitySetsService(
...
+            && this::propertyTypes.isInitialized
+
+    fun initializeOrganizationMetadataEntitySets(organizationId: UUID) {
+        val adminAclKey = organizations.executeOnKey(
+            organizationId,
+            OrganizationReadEntryProcessor { it.adminRoleAclKey }
+        ) as AclKey
+
+
+        initializeOrganizationMetadataEntitySets(securePrincipalsManager.getRole(organizationId, adminAclKey[1]))
```
katherinebernstein, line 116: This should probably be using `PrincipalsMapManager.lookupRole` (as PrincipalsMapManager is intended to be a lightweight read layer over the mapstore and `SecurePrincipalsManager` is intended for more heavy-duty/workflow-specific computation on top of that) -- and that would also let you just pass `adminAclKey` directly instead of having to do hacky deconstruction of the AclKey

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -254,25 +284,29 @@ class OrganizationMetadataEntitySetsService(
...
-                    propertyTypes.getValue(DATASET_NAME).id to setOf(table.name),
-                    propertyTypes.getValue(STANDARDIZED).id to setOf(false),
-                    propertyTypes.getValue(COL_INFO).id to setOf(mapper.writeValueAsString(tableColumnEntities.values))
+                propertyTypes.getValue(DATASET_NAME).id to setOf(table.name),
+                propertyTypes.getValue(DESCRIPTION).id to setOf(table.description),
+                propertyTypes.getValue(ID).id to setOf(table.id.toString()),
+                propertyTypes.getValue(ORG_ID).id to setOf(organizationId),
+                propertyTypes.getValue(PG_OID).id to setOf(table.oid),
+                propertyTypes.getValue(STANDARDIZED).id to setOf(false),
+                propertyTypes.getValue(TITLE).id to setOf(table.title)
```
katherinebernstein, line 308: we should probably add a property type + store schema here too

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -293,13 +327,17 @@ class OrganizationMetadataEntitySetsService(
...
-                propertyTypes.getValue(ID).id to setOf(table.id.toString()),
-                propertyTypes.getValue(DATASET_NAME).id to setOf(table.name),
-                propertyTypes.getValue(STANDARDIZED).id to setOf(false)
+            propertyTypes.getValue(DATASET_NAME).id to setOf(table.name),
+            propertyTypes.getValue(DESCRIPTION).id to setOf(table.description),
+            propertyTypes.getValue(ID).id to setOf(table.id.toString()),
+            propertyTypes.getValue(ORG_ID).id to setOf(organizationId),
+            propertyTypes.getValue(PG_OID).id to setOf(table.oid),
+            propertyTypes.getValue(STANDARDIZED).id to setOf(false),
+            propertyTypes.getValue(TITLE).id to setOf(table.title)
```
katherinebernstein, line 340: is this the same entity mapping as above? if yes we should probably extract this into its own like `mapExternalTableToEntity` method to avoid copy/paste issues, especially while the data model here is still developing (assuming we're sticking with the metadata entity set approach, which I have separate opinions about)

katherinebernstein: and same for all the other entity mappings -- column, property type, entity set, etc

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -413,16 +448,34 @@ class OrganizationMetadataEntitySetsService(
...
+            _description = "Dataset columns for $organizationId",
             contacts = mutableSetOf(),
             flags = EnumSet.of(EntitySetFlag.METADATA)
     )
 
-    private fun buildOrganizationMetadataEntitySetName(organizationId: UUID): String = quote(
-            "org-metadata-$organizationId"
-    )
-
+    private fun buildOrganizationMetadataEntitySetName(organizationId: UUID): String = quote("org-metadata-$organizationId")
```
katherinebernstein, line 457: why is this wrapped in quotes?

katherinebernstein: (same question for the other entity set names)

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -334,17 +382,6 @@ class OrganizationMetadataEntitySetsService(
                 columnEntities,
                 columnAuthorizedPropertyTypes
         )
-
-        val datasetEntityKeyId = getDatasetEntityKeyId(organizationMetadataEntitySetIds, table.id)
-        val datasetColumnEntity = mutableMapOf<UUID, Set<Any>>(
-                propertyTypes.getValue(COL_INFO).id to setOf(mapper.writeValueAsString(columnEntities))
-        )
-
```
katherinebernstein, line 342: what's the reason for deleting this chunk?

UnsungHero97: `ol.columninfo` is no longer needed so it doesn't need to exist.

katherinebernstein: so are we not supporting keyword searches executing across tables + columns anymore?

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
UnsungHero97: I don't know. Git history shows the quotes have been there for since the monorepo, so I assume they've been there since this class was introduced.

katherinebernstein: hm I think that's a bug actually -- this will cause the entity set names to be stored with quotes as part of the name, which is definitely not intended and is likely to cause issues with a bunch of external tools. If the name is only used for entity set initialization, you should probably remove that.

makosblade:
So I am a little confused by the broader pattern here. It seems as though the intent of OrganizationMetadataEntitySets is to store metadata about organizations, specifically external table information, in EntitySets. I'm not sure I follow why this is hte better choice over enhancing the ExternalDatabaseManagement codepaths

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
UnsungHero97: I prefer to do that in a follow-up pull request.

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
UnsungHero97: It does look weird to me, happy to remove if you think it's a bug.

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
UnsungHero97: Keyword search still works for tables and columns. Are you asking if we are supporting a single search request for both tables and columns at the same time?

UnsungHero97:
@makosblade, my goal with this pull request was to extract what I need from https://github.com/openlattice/openlattice/pull/88 in order to unblock the frontend.

It may very well be the case that `ExternalDatabaseManagementService` is a better place to handle the metadata logic, I just haven't looked at it in detail to have an answer for you. One note is that `OrganizationMetadataEntitySetsService` is handling both external tables as well as regular entity sets, the idea being that the `datasets` metadata entity set contains ALL data sets that an organization might have, be it internal entity sets for apps, external tables in atlas, external tables in snowflake, etc.

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
katherinebernstein: yeah that would be ideal especially if you're recreating them all -- it is a bug.

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
katherinebernstein: yeah, seems like at the very least this will break paging.

makosblade:
> @makosblade, my goal with this pull request was to extract what I need from #88 in order to unblock the frontend.
> 
> It may very well be the case that `ExternalDatabaseManagementService` is a better place to handle the metadata logic, I just haven't looked at it in detail to have an answer for you. One note is that `OrganizationMetadataEntitySetsService` is handling both external tables as well as regular entity sets, the idea being that the `datasets` metadata entity set contains ALL data sets that an organization might have, be it internal entity sets for apps, external tables in atlas, external tables in snowflake, etc.

Okay, I see. The current `OrganizationExternalTables` tracking includes all of that as well so I think we really need to decide which direction we're going in here long term since this work duplicates a bunch of existing patterns. I'm fine with merging this to unblock you for CA pending @katherinebernstein 's comments but let's definitely revisit it afterwards.

katherinebernstein (approved):

makosblade (approved):

PR #122, opened 2021-02-22
grant admin role to user creator of org

katherinebernstein:


makosblade (approved):

datastore/src/main/kotlin/com/openlattice/organizations/controllers/OrganizationsController.kt
```diff
@@ -109,15 +109,10 @@ class OrganizationsController : AuthorizingComponent, OrganizationsApi {
...
         Preconditions.checkArgument(
-                organization.connections.isEmpty() || isAdmin(),
+                organization.connections.isEmpty() || isAdmin,
                 "Must be admin to specify auto-enrollments"
         )
-        organizations.createOrganization(Principals.getCurrentUser(), organization)
-        securableObjectTypes.createSecurableObjectType(
-                AclKey(organization.id),
-                SecurableObjectType.Organization
-        )
```
UnsungHero97, line 119: Why does this not need to exist anymore?

UnsungHero97 (approved):

datastore/src/main/kotlin/com/openlattice/organizations/controllers/OrganizationsController.kt
katherinebernstein: it never did. this shouldn't even be in the controller anyway.

PR #131, opened 2021-03-04
allow org members to remove themselves

UnsungHero97:
At the moment, only users with OWNER on the org are allowed to add / remove members. As such, an org member cannot "leave" the org. I think org members should be allowed to leave on their own discretion.

The changes in the api file are just formatting changes.

### DIFF
--- a/datastore/src/main/kotlin/com/openlattice/organizations/controllers/OrganizationsController.kt
+++ b/datastore/src/main/kotlin/com/openlattice/organizations/controllers/OrganizationsController.kt
@@ -537,12 +537,11 @@ class OrganizationsController : AuthorizingComponent, OrganizationsApi {
             @PathVariable(OrganizationsApi.ID) organizationId: UUID,
             @PathVariable(OrganizationsApi.USER_ID) userId: String
     ): Void? {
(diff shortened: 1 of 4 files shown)


datastore/src/main/kotlin/com/openlattice/organizations/controllers/OrganizationsController.kt
```diff
@@ -537,12 +537,12 @@ class OrganizationsController : AuthorizingComponent, OrganizationsApi {
...
     ): Void? {
-        ensureOwnerAccess(AclKey(organizationId))
-        organizations.removeMembers(
-                organizationId, ImmutableSet.of(
-                Principal(PrincipalType.USER, userId)
-        )
-        )
+        val principal = Principal(PrincipalType.USER, userId)
+        // allow caller to remove themselves from the org
+        if (Principals.getCurrentPrincipals().contains(principal).not()) {
```
katherinebernstein, line 542: It's much lighter and also safer to use this check instead
```
if (Principals.getCurrentUser().id == userId) { ... }
```

katherinebernstein (approved):

makosblade (approved):

PR #137, opened 2021-03-15
Bugfix/data deletion jobs

katherinebernstein:


conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJob.kt
```diff
@@ -0,0 +1,524 @@
...
+
+    private fun cleanUpBatch(entityDataKeys: Set<EntityDataKey>) {
+        val entityKeyIds = entityDataKeys.mapTo(mutableSetOf()) { it.entityKeyId }
+        PostgresLinkingQueryService.deleteNeighborhoods(hds, state.entitySetId, entityKeyIds)
+
+        if (state.entityKeyIds == null) {
+            return
+        }
+
+        state.entityKeyIds!!.removeAll(entityKeyIds)
```
makosblade, line 161: I don't think you need the `!!` here 🤔

makosblade: and on 143 as well

makosblade (dismissed):

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJob.kt
katherinebernstein: They are needed (I think since the state.entityKeyIds value could, in theory, change in between these calls) but I can make this more kotliny

makosblade (approved):

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -756,14 +787,7 @@ public Long deleteEntitiesAndNeighbors(
...
 
-        WriteEvent writeEvent = deletionManager.clearOrDeleteEntitiesAndNeighborsIfAuthorized(
-                entitySetId,
-                filter.getEntityKeyIds(),
-                filter.getSrcEntitySetIds().orElse( ImmutableSet.of() ),
-                filter.getDstEntitySetIds().orElse( ImmutableSet.of() ),
-                deleteType,
-                Principals.getCurrentPrincipals()
-        );
+        WriteEvent writeEvent = new WriteEvent( 0, 0 );
```
UnsungHero97, line 790: From what I can tell, `deleteEntitiesAndNeighbors` no longer does anything. Is this intentional?

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -1059,4 +1083,32 @@ private void ensureEntitySetsCanBeWritten( Set<UUID> entitySetIds ) {
         }
     }
 
+    private int blockOnDeletionJobGettingNumUpdates( UUID deletionJobId, boolean blockUntilCompletion ) {
+        if ( !blockUntilCompletion ) {
+            return 0;
+        }
+
+        for ( int i = 0; i < MAX_DELETION_BLOCKING_CHECKS; i++ ) {
```
UnsungHero97, line 1091: `MAX_DELETION_BLOCKING_CHECKS` is defined as 

```
MAX_DELETION_BLOCKING_CHECKS = DELETION_BLOCKING_INTERVAL * 60 * 5; // 5 minutes
```

... which means the for-loop will run for 300,000 iterations. Since there's a `Thread.sleep` for 1 second, doesn't that mean the for-loop should only do 300 iterations? 

In other words, shouldn't `i++` actually be `i += 1000`?

indexer/src/main/kotlin/com/openlattice/indexing/BackgroundIndexedEntitiesDeletionService.kt
```diff
@@ -187,4 +245,10 @@ class BackgroundIndexedEntitiesDeletionService(
     private fun updateExpiration(entitySetId: UUID) {
         deletionLocks.set(entitySetId, System.currentTimeMillis() + DELETE_EXPIRATION_MILLIS)
     }
+
+    data class EntitySetForDeletion(
+            val id: UUID,
+            val name: String,
+            val partitions: Set<Int>
+    )
```
UnsungHero97, line 253: This should be in its own file.

indexer/src/main/kotlin/com/openlattice/indexing/BackgroundIndexedEntitiesDeletionService.kt
```diff
@@ -92,47 +98,65 @@ class BackgroundIndexedEntitiesDeletionService(
             val w = Stopwatch.createStarted()
             //We shuffle entity sets to make sure we have a chance to work share and index everything
             val lockedEntitySets = entitySets.values
+                    .map { EntitySetForDeletion(it.id, it.name, it.partitions) }
```
UnsungHero97, line 101: This was changed very recently ... just flagging this to double check with @makosblade

indexer/src/main/kotlin/com/openlattice/indexing/BackgroundIndexedEntitiesDeletionService.kt
```diff
@@ -92,47 +98,65 @@ class BackgroundIndexedEntitiesDeletionService(
...
                     .shuffled()
-                    .filter { it.name != "OpenLattice Audit Entity Set" } //TODO: Clean out audit entity set from prod
                     .filter { tryLockEntitySet(it.id) }
-                    .map { IndexerEntitySetMetadata.fromEntitySet(it) }
 
-            val totalDeleted = lockedEntitySets
+            val lockedDeletedEntitySets = deletedEntitySets
+                    .map { (id, partitions) -> EntitySetForDeletion(id, "Deleted entity set [$id]", partitions) }
```
UnsungHero97, line 106: What is this magic string `"Deleted entity set [$id]"` ?

api/src/main/java/com/openlattice/data/DataApi.java
```diff
@@ -229,7 +229,8 @@ Integer deleteEntity(
     Integer deleteEntities(
             @Path( ENTITY_SET_ID ) UUID entitySetId,
             @Body Set<UUID> entityKeyIds,
-            @Query( TYPE ) DeleteType deleteType );
+            @Query( TYPE ) DeleteType deleteType,
+            @Query( BLOCK ) boolean blockUntilCompletion );
```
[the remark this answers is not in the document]
katherinebernstein: this is a pretty common endpoint for apps to call so I didn't want to change the behavior there. `blockUntilCompletion` is a bit hacky but is true by default which will ensure that the apps don't break when this is released.

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJobState.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.geekbeast.rhizome.jobs.JobState
+import com.openlattice.data.DeleteType
+import java.util.*
+
+data class DataDeletionJobState(
+        val entitySetId: UUID,
+        val deleteType: DeleteType,
+        var partitions: Set<Int>,
+        val entityKeyIds: MutableSet<UUID>? = null,
```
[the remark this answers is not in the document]
katherinebernstein: null and empty set have different meanings. not providing entityKeyIds is logically more consistent with not narrowing the scope of the delete to specific entityKeyIds.

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJob.kt
```diff
@@ -0,0 +1,521 @@
...
+            state: DataDeletionJobState
+    ) : this(state) {
+        initialize(id, taskId, status, progress, hasWorkRemaining, result)
+    }
+
+    override val resumable = true
+
+    companion object {
+        private const val BATCH_SIZE = 10_000
+        private val ALL_PARTITIONS = (0..257)
```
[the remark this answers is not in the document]
katherinebernstein: yeah there was a bug a while back that allowed partition 257 to exist, and there's still one entity set that has it.

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJobState.kt
```diff
@@ -0,0 +1,14 @@
...
+        internal var totalToDelete: Long = 0,
+        var numDeletes: Long = 0
```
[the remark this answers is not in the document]
katherinebernstein: no. it's used ouside of DataDeletionJob.

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -718,14 +733,30 @@ public Integer deleteEntityProperties(
             @PathVariable( ENTITY_KEY_ID ) UUID entityKeyId,
             @RequestBody Set<UUID> propertyTypeIds,
             @RequestParam( value = TYPE ) DeleteType deleteType ) {
-
+        if ( deleteType.equals( DeleteType.Soft ) ) {
+            ensureWriteAccess( new AclKey( entitySetId ) );
+        } else {
+            ensureOwnerAccess( new AclKey( entitySetId ) );
+        }
```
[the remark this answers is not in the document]
katherinebernstein: was avoiding that extra pointer but this is fine

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
katherinebernstein: yes

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
katherinebernstein: ah yeah good catch!

indexer/src/main/kotlin/com/openlattice/indexing/BackgroundIndexedEntitiesDeletionService.kt
katherinebernstein: nope actually since it's a data class that exists solely within this class solely as a helper for extracting certain features and doesn't have any intrinsic meaning, it should exist only in this class.

indexer/src/main/kotlin/com/openlattice/indexing/BackgroundIndexedEntitiesDeletionService.kt
katherinebernstein: we don't store names for deleted entity sets.

katherinebernstein: (this string is what will print in the logs instead of the name)

indexer/src/main/kotlin/com/openlattice/indexing/BackgroundIndexedEntitiesDeletionService.kt
katherinebernstein: yeah I didn't change anything about the recent pattern, EntitySetForDeletion is effectively the same as IndexerEntitySetMetadata minus entityTypeId to accommodate deleted entity sets.

UnsungHero97 (approved):

makosblade (approved):

PR #138, opened 2021-03-27
Task/add reentry email template

giuliacampana:
adds a separate Reentry Case Management task alert email template and renderer (separate from the CARE one) because Reentry has slightly different fields than CARE, and of course a different app URL.

### DIFF
--- a/conductor-client/src/main/kotlin/com/openlattice/search/renderers/ReentryTaskAlertEmailRenderer.kt
+++ b/conductor-client/src/main/kotlin/com/openlattice/search/renderers/ReentryTaskAlertEmailRenderer.kt
@@ -50,11 +50,20 @@ class ReentryTaskAlertEmailRenderer {
     companion object {
 
         private fun isMatchByEntitySetId(neighbor: NeighborEntityDetails, entitySetId: UUID, associationEntitySetId: Optional<UUID>): Boolean {
-            val isMatchingEntitySetId = neighbor.neighborEntitySet.isPresent && neighbor.neighborEntitySet.get().id == entitySetId
-            return if (associationEntitySetId.isPresent) {
-                (neighbor.associationEntitySet.id == associationEntitySetId.get()) && isMatchingEntitySetId
+
+            if (neighbor.neighborEntitySet.isEmpty) {
+              return false
             }
-            else { isMatchingEntitySetId }
+
+            if (neighbor.neighborEntitySet.get().id != entitySetId) {
+              return false
+            }
+
+            if (associationEntitySetId.isPresent && associationEntitySetId.get() != neighbor.associationEntitySet.id) {
+              return false
+            }
+
+            return true
         }
 
         private fun getCombinedNeighbors(
--- a/conductor-client/src/main/kotlin/com/openlattice/search/renderers/ReentryTaskAlertEmailRenderer.kt
+++ b/conductor-client/src/main/kotlin/com/openlattice/search/renderers/ReentryTaskAlertEmailRenderer.kt
@@ -1,0 +1,165 @@
+package com.openlattice.search.renderers
+
+import com.google.common.collect.Maps
+import com.openlattice.data.requests.NeighborEntityDetails
+import com.openlattice.edm.EdmConstants
+import com.openlattice.mail.RenderableEmailRequest
+import com.openlattice.search.requests.PersistentSearch
+import org.apache.olingo.commons.api.edm.FullQualifiedName
+import java.time.OffsetDateTime
+import java.util.*
+
+
+private const val FROM_EMAIL = "courier@openlattice.com"
+private const val TEMPLATE_PATH = "mail/templates/shared/CAREIssueAlertTemplate.mustache"
+
+/* FQNs */
+private val PERSON_ID_FQN = FullQualifiedName("nc.SubjectIdentification")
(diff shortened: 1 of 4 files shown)


conductor-client/src/main/resources/mail/templates/shared/ReentryTaskAlertTemplate.mustache
```diff
@@ -0,0 +1,79 @@
+<!doctype html>
+<html>
+<head>
+    <title>Reentry Case Management - Task{{subtitle}}</title>
```
UnsungHero97, line 4: `Task{{subtitle}}` won't this mean the `Task` and whatever follow are right next to each other? Is that intentional?

conductor-client/src/main/resources/mail/templates/shared/ReentryTaskAlertTemplate.mustache
```diff
@@ -0,0 +1,79 @@
...
+        line-height: normal;
+        color: #8e929b;
+        padding-top: 10px;
+      }
+
+      body {
+        font-family: 'Inter', Arial, sans-serif;
+        line-height: 1.5;
+        color: #333;
+      }
```
UnsungHero97, line 30: This might not actually matter, but please put body at the top.

conductor-client/src/main/resources/mail/templates/shared/ReentryTaskAlertTemplate.mustache
```diff
@@ -0,0 +1,79 @@
...
+
+      .divider {
+        width: 100%;
+        background-color: #e1e1eb;
+        height: 1px;
+        margin: 20px 0;
+      }
+
+      .detail {
+        font-family: 'Inter', Arial, sans-serif;
```
UnsungHero97, line 40: I don't think you need to specify this since it's already set on the body.

conductor-client/src/main/resources/mail/templates/shared/ReentryTaskAlertTemplate.mustache
```diff
@@ -0,0 +1,79 @@
...
+        padding-top: 5px;
+      }
+
+      .footer {
+        padding-top: 10px;
+        border-top: 1px solid #e1e1eb;
+        margin-top: 30px;
+        color: #717175;
```
UnsungHero97, line 48: Please sort.

UnsungHero97 (changes requested):

conductor-client/src/main/resources/mail/templates/shared/ReentryTaskAlertTemplate.mustache
giuliacampana: ah ya, it is actually intentional. if there is a subtitle on the entity, then this subtitle variable will look like ``: ${subtitle}``.

conductor-client/src/main/kotlin/com/openlattice/search/renderers/ReentryTaskAlertEmailRenderer.kt
```diff
@@ -0,0 +1,165 @@
...
+            val dueDateTime = (issue[GENERAL_DATETIME_FQN] ?: emptySet()).map { OffsetDateTime.parse(it.toString()) }
+            val dueDate = dueDateTime.joinToString(", ") { MessageFormatters.formatDate(it, timeZone) }
+            val dueTime = completedDateTime.joinToString(", ") { MessageFormatters.formatTime(it, timeZone) }
+
+            return mapOf(
+                    "dueDateTime" to "$dueDate $dueTime",
+                    "status" to (issue[STATUS_FQN] ?: emptySet()).joinToString(", "),
+                    "title" to (issue[TITLE_FQN] ?: emptySet()).joinToString(", "),
+                    "description" to (issue[DESCRIPTION_FQN] ?: emptySet()).joinToString(", "),
+                    "category" to (issue[CATEGORY_FQN] ?: emptySet()).joinToString(", ")
```
anzioka, line 114: probably a good idea to constantify these and other strings

UnsungHero97 (approved):

anzioka (approved):

conductor-client/src/main/kotlin/com/openlattice/search/renderers/ReentryTaskAlertEmailRenderer.kt
```diff
@@ -0,0 +1,171 @@
...
+*   "timezone": <TimeZones>,
+* }
+*
+*/
+
+class ReentryTaskAlertEmailRenderer {
+
+    companion object {
+
+        private fun isMatchByEntitySetId(neighbor: NeighborEntityDetails, entitySetId: UUID, associationEntitySetId: Optional<UUID>): Boolean {
```
katherinebernstein, line 52: This code is totally valid and is all good as is -- I might restructure it a little bit just for clarity though. Something like
```
if (neighbor.neighborEntitySet.isEmpty) {
  return false
}

if (neighbor.neighborEntitySet.get().id != entitySetId) {
  return false
}

if (associationEntitySetId.isPresent && associationEntitySetId.get() != neighbor.associationEntitySet.id) {
  return false
}

return true
```

katherinebernstein (approved):

anzioka (approved):

UnsungHero97 (approved):

PR #150, opened 2021-04-06
update org metadata service with deleteDatasets and deleteDatasetColumns

UnsungHero97:


conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -254,7 +257,7 @@ class OrganizationMetadataEntitySetsService(
         val organizationMetadataEntitySetIds = organizationService.getOrganizationMetadataEntitySetIds(organizationId)
         val datasetEntityKeyIds = getDatasetEntityKeyIds(
             organizationMetadataEntitySetIds,
-            tables.map { it.id }
+            tables.map { it.oid }
```
katherinebernstein, line 260: this is going to break all existing data and duplicate everything (within the metadata entity sets) -- is that intentional?

katherinebernstein: oids also will also be reused, so that means table data will be written to the same entities as old deleted data, which seems bad...

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
```diff
@@ -209,27 +209,37 @@ class BackgroundExternalDatabaseSyncingService(
...
+        existingTableIds: Set<UUID>,
+        existingColumnIds: Set<UUID>
+    ) {
 
         // delete missing tables
 
-        val tableIdsToDelete = organizationExternalDatabaseTables.keySet(Predicates.equal(ORGANIZATION_ID_INDEX, orgId)).filter {
-            !existingTableIds.contains(it)
-        }.toSet()
+        val tablesToDelete: Set<OrganizationExternalDatabaseTable> = organizationExternalDatabaseTables
```
katherinebernstein, line 220: why are you loading the full table objects here?

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
```diff
@@ -209,27 +209,37 @@ class BackgroundExternalDatabaseSyncingService(
...
+            .values(Predicates.equal(ORGANIZATION_ID_INDEX, orgId))
+            .filter { !existingTableIds.contains(it.id) }
+            .toSet()
+        val tableIdsToDelete = tablesToDelete.mapTo(mutableSetOf<UUID>()) { it.id }
 
         if (tableIdsToDelete.isNotEmpty()) {
             edms.deleteOrganizationExternalDatabaseTableObjects(tableIdsToDelete)
+            organizationMetadataEntitySetsService.deleteDatasets(orgId, tablesToDelete)
```
katherinebernstein, line 228: imo i think it would be a bit cleaner/safer to add this logic as part of the existing delete external tables workflow (within ExternalDatabaseManagementService) but it doesn't really matter

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
UnsungHero97: It is not intentional to break all existing data and duplicate everything. The issue is that the table `oid` is currently in use in prod and I thought this change is least dangerous.

In this pull request, `addDataset` is calling `getDatasetEntityKeyId` with `table.oid`:

https://github.com/openlattice/openlattice/pull/76/files#diff-98470d2cb0c12678b30beee4b37ffcaec7dc44416b0a553c6cbceb89335f1256R305

However, `addDatasetsAndColumns` is calling `getDatasetEntityKeyIds` with `table.id`:

https://github.com/openlattice/openlattice/pull/76/files#diff-98470d2cb0c12678b30beee4b37ffcaec7dc44416b0a553c6cbceb89335f1256R244

Unfortunately, `addDataset` is being called by `BackgroundExternalDatabaseSyncingService` here:

https://github.com/openlattice/openlattice/blob/develop/indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt#L183

As far as I can tell, this means that all new tables that are being synced right now are reserving ekids using the table `oid` instead of the `id`, which I've confirmed by looking in the `sync_ids` table.

On the other hand, `addDatasetsAndColumns` is using the table `id`, which sounds like is the correct logic. However, `addDatasetsAndColumns` is only called by `importMetadata` in `OrganizationsController`, which is not really used anywhere as far as I know.

Alternatively, I can undo this change and update `addDataset` to use the table `id`, though I think that would require a migration.

Do you have another recommendation?

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
UnsungHero97: As it's currently implemented, I need to pass the table objects to `deleteDatasets` below because I need to access the table `oid`. This will probably change depending on what we decide to do with the `oid` issue in the discussion above.

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
UnsungHero97: I thought about it, but decided against it because I wanted to avoid making changes that will have to be undone since the plan is to nuke `OrganizationMetadataEntitySetsService` anyway.

conductor-client/src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
katherinebernstein: I think we should definitely be using id for entityKeyId over oid. If there's a spot oid has been used in prod, we'll need to wipe that data and reintegrate because that means we have a ton of duplicates.

UnsungHero97: Ok, I'll make the update in this pull request and will move the data conversation to slack.

indexer/src/main/kotlin/com/openlattice/BackgroundExternalDatabaseSyncingService.kt
UnsungHero97: Fixed.

katherinebernstein (approved):

makosblade (approved):

PR #153, opened 2021-04-09
delete a whole bunch of unused code, fix a bunch of warnings

katherinebernstein:


conductor-client/src/main/java/com/openlattice/organizations/roles/SecurePrincipalsManager.java
```diff
@@ -128,18 +116,6 @@ void createSecurablePrincipal(
...
-    /**
-     * Returns all Principals, which have all the specified permissions on the securable object
-     *
-     * @param key         The securable object
-     * @param permissions Set of permission to check for
-     */
-    Set<Principal> getAuthorizedPrincipalsOnSecurableObject( AclKey key, EnumSet<Permission> permissions );
-
-    SecurablePrincipal getSecurablePrincipalById( UUID id );
-
```
makosblade, line 142: I get that these are not currently used but a couple of them seem helpful to have around, `getAuthorizedPrincipalsOnSecurableObject` and `getSecurablePrincipalById`. 
Are there other calls that effectively replace the need for these two?

api/src/main/java/com/openlattice/search/requests/Search.java
```diff
@@ -46,7 +46,7 @@ public Search(
             @JsonProperty( SerializationConstants.EXCLUDE_PROPERTY_TYPES ) Optional<Boolean> excludePropertyTypes,
             @JsonProperty( SerializationConstants.START ) int start,
             @JsonProperty( SerializationConstants.MAX_HITS ) int maxHits ) {
-        optionalKeyword = keyword.isPresent() ? Optional.of( keyword.get().trim() ) : Optional.empty();
+        optionalKeyword = keyword.map( String::trim );
```
makosblade, line 49: Does this end up keeping it Optional.empty if it is absent? I didn't even knew there was a map for optional

api/src/main/java/com/openlattice/edm/requests/MetadataUpdate.java
```diff
@@ -73,14 +78,14 @@ public MetadataUpdate(
...
-        Preconditions.checkArgument( !name.isPresent() || StringUtils.isNotBlank( name.get() ),
+        Preconditions.checkArgument( name.isEmpty() || StringUtils.isNotBlank( name.get() ),
                 "Entity set name cannot be blank." );
-        Preconditions.checkArgument( !contacts.isPresent() || !contacts.get().isEmpty(), "Contacts cannot be blank." );
-        Preconditions.checkArgument( !type.isPresent() || StringUtils.isNotBlank( type.get().getNamespace() ),
+        Preconditions.checkArgument( contacts.isEmpty() || !contacts.get().isEmpty(), "Contacts cannot be blank." );
+        Preconditions.checkArgument( type.isEmpty() || StringUtils.isNotBlank( type.get().getNamespace() ),
                 "Namespace of type is missing." );
-        Preconditions.checkArgument( !type.isPresent() || StringUtils.isNotBlank( type.get().getName() ),
+        Preconditions.checkArgument( type.isEmpty() || StringUtils.isNotBlank( type.get().getName() ),
```
makosblade, line 88: Another way I've been cleaning these up lately is using `require`, `requireNotNull`, `check`, and `checkNotNull`. Lets you perform an action before the exception is thrown so you can `logger` before it gets thrown. Also feels cleaner than `Preconditions.checkArgument`

makosblade (commented):
I just have a couple more files to look through but looks good so far. Will come back to it later today

conductor-client/src/main/java/com/openlattice/organizations/roles/SecurePrincipalsManager.java
katherinebernstein: Yep -- getAuthorizedPrincipalsOnSecurableObject is in AuthorizationsManager (better there than PrincipalsManager), and getSecurablePrincipalById moved to PrinicpalMapManager (and is by AclKey, not id)

api/src/main/java/com/openlattice/search/requests/Search.java
katherinebernstein: yep! I didn't know about this one either

api/src/main/java/com/openlattice/edm/requests/MetadataUpdate.java
katherinebernstein: I thought that was only available in Kotlin

makosblade: Oh derrrp. Yep didn't realize this was Java

makosblade (approved):
LGTM, sounds like the `*Graph*` stuff is completely unused and unrelated to `GraphApi` paths which, IMO, may eventually be re-used

UnsungHero97 (approved):

api/src/main/java/com/openlattice/edm/requests/EdmRequest.java
```diff
@@ -18,19 +18,19 @@
...
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.openlattice.client.serialization.SerializationConstants;
+
+import java.util.Set;
+import java.util.UUID;
 
 public class EdmRequest {
-    public static enum Action {
+    public enum Action {
```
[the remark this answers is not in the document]
katherinebernstein: that's definitely an option, but it will involve designing what REPLACE means in the context of a permission request or otherwise handling it being passed as transforming it to SET which feels like bad practice but the correct behavior. The best thing is probably to get rid of this class altogether since I'm pretty sure its usage is in super highly deprecated, never used, extremely broken APIs that we can probably just delete, but I was trying to steer clear of actually deleting APIs in this PR and stick to more straightforward deletes.

PR #154, opened 2021-04-10
improvements to background entity deletion, involving locking, log le…

katherinebernstein:
…vel, renames for clarity, and a batch issue


conductor-client/src/main/kotlin/com/openlattice/postgres/PostgresLockingUtils.kt
```diff
@@ -132,49 +131,6 @@ fun lockIdsAndExecute(
     }
 }
 
-fun lockIdsAndExecuteAndCommit(
```
makosblade, line 135: This is still in use by linker

makosblade: nvm, this is a very similarly named function lol

makosblade (approved):

UnsungHero97 (approved):

PR #170, opened 2021-04-23
search users using fields

UnsungHero97:


api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
```diff
@@ -0,0 +1,21 @@
+package com.openlattice.search
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.util.Optional
+
+// https://auth0.com/docs/users/user-search/user-search-query-syntax
+data class Auth0UserSearchFields(
+    @JsonProperty val email: Optional<String> = Optional.empty(),
+    @JsonProperty val name: Optional<String> = Optional.empty()
```
katherinebernstein, line 9: is there a reason you're using optionals here instead of just defaulting to null?

conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
```diff
@@ -42,28 +42,45 @@ class Auth0UserDirectoryService(
...
+        // https://auth0.com/docs/users/user-search/user-search-query-syntax
+        // TODO - support multiple fields and construct a valid Lucene query string to pass to Auth0
+        if (fields.email.orElse("").isNotBlank()) {
+            searchQuery = "email:${fields.email.get()}"
         }
+        else if (fields.name.orElse("").isNotBlank()) {
+            searchQuery = "name:${fields.name.get()}"
+        }
+
+        require(searchQuery.isNotBlank()) { "search query cannot be blank" }
```
katherinebernstein, line 58: these checks are unnecessary since blank states are impossible because of constructor assertions. simpler/cleaner to just do 
```
val searchQuery = if (fields.name.isPresent) {
  "name:${fields.name.get()}"
} else {
  "email:${fields.email.get()}"
}
```

(or with null checks and no .get()s if you get rid of the optionals)

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
UnsungHero97: Yea... this is just a symptom of me not being as familiar/comfortable with Kotlin, but dealing with the nulls got messy earlier on in the development cycle. I can try switching back if you'd like.

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
```diff
@@ -0,0 +1,20 @@
+package com.openlattice.search
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+// https://auth0.com/docs/users/user-search/user-search-query-syntax
+data class Auth0UserSearchFields(
+    @JsonProperty val email: String?,
+    @JsonProperty val name: String?
```
katherinebernstein, line 8: so you probably do still want the default values here so that users can pass
```
{
  "email": "blah"
}
```
instead of 
```
{
  "email": "blah",
  "name": null
}
```
(or vice versa)

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
```diff
@@ -0,0 +1,20 @@
...
+) {
+    init {
+        // TODO - support multiple fields and construct a valid Lucene query string to pass to Auth0
```
makosblade, line 11: Can you create a JIRA ticket and add it here?
ex: `TODO - JIRA-XXXX - `

conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
```diff
@@ -42,28 +42,43 @@ class Auth0UserDirectoryService(
...
-            users.addAll(pageOfUsers)
+        var searchQuery = ""
 
-            if (pageOfUsers.size == DEFAULT_PAGE_SIZE) {
-                pageOfUsers = auth0ManagementApi.searchAllUsers(searchQuery, page++, DEFAULT_PAGE_SIZE, "v3")
-            } else {
-                break
-            }
+        // https://auth0.com/docs/users/user-search/user-search-query-syntax
+        // TODO - support multiple fields and construct a valid Lucene query string to pass to Auth0
```
makosblade, line 50: Same as above re: JIRA ticket

conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
```diff
@@ -42,28 +42,43 @@ class Auth0UserDirectoryService(
     }
 
     //TODO: Switch over to a Hazelcast map to relieve pressure from Auth0
```
makosblade, line 44: Just flagging for follow-up, anyone know what the idea here is? 
I'm guessing a mapstore that is backed by querying Auth0?

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
```diff
@@ -0,0 +1,20 @@
...
+        require((email != null && name != null).not()) { "only one of \"email\", \"name\" are allowed" }
```
katherinebernstein, line 12: I think there's a missing assertion here that at least one of them must be present. I think it's a bit easier to read this as
```
require(email != null || name != null) { "<can't both be null, blah>" } 
require(email == null || name == null) { "<can't both be present, blah>" }

...not blank assertions...
```

or! if you want to be super fancy and do all these constrictor assertion in one line, you could do:
```
require(arrayOf(name, email).filterNot { it.isNullOrBlank() }.size == 1) {
  <exactly one of email, name must be present and not blank, blah>
}
```

(although the former is way more intuitive so I prefer it for that reason)

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
```diff
@@ -0,0 +1,20 @@
...
+        if (email != null) {
+            require(email.isNotBlank()) { "email cannot be blank" }
+        }
+        if (name != null) {
+            require(name.isNotBlank()) { "name cannot be blank" }
+        }
+    }
```
makosblade, line 19: Looks like the use case is to have only one or the other field filled but not both. 
One way I like of doing that is as follows using factory methods:

```
class Auth0UserSearchFields private constructor( val email: String?, val name: Int? ) {
    companion object {
        @JvmStatic
        fun emailSearch( val email ): Auth0UserSearchFields {
            return Auth0UserSearchFields(email, null)
        }
        
        @JvmStatic
        fun nameSearch( val name ): Auth0UserSearchFields {
            return Auth0UserSearchFields(null, name)
        }
    }
}
```

That way the user of this class never has the option of providing an invalid search that has both name and email because the constructor is private

makosblade (approved):
Just a couple of little documentation things and some food for thought

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
UnsungHero97: Passing `null` for `name` is not necessary. This works as expected:

```
{
  "email": "blah"
}
```

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
UnsungHero97: I just caught this bug too. I'll upate.

conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
UnsungHero97: This has been around for over a year, not sure what the intention is, though "a mapstore backed by querying Auth0" makes sense to me.

https://github.com/openlattice/conductor-client/pull/749/files#diff-52f449f6251239dda152342be19580a17f4bee38b4ba62b854b2944eae3f8bf4R35

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
katherinebernstein: cool cool. ah the other thing is that you don't need to pass `@JsonProperty` -- Jackson knows how to properly serialize/deserialize data classes

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
UnsungHero97: I like it, but it's not what I was going for. The query Auth0 accepts is a Lucene query so it's totally possible to do the following:

```
name:test* AND email:test@openlattice.com
```

The issue is basically that I'm not sure where the "Lucene query builder" logic belongs ... is it on the frontend? ... is it on the backend? ... is it a mixture of both? I don't know. Additionally, we can search other Auth0 fields like `given_name`, `family_name, `email_verified`, etc. While we don't have a use case for that at the moment, I thought it would be easier to add support for additional fields in the future if we have a simple data class.

conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
katherinebernstein: I assume this was from before the users mapstore existed, which now exists, so this can probably be deleted

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
UnsungHero97: https://jira.openlattice.com/browse/LATTICE-2805

conductor-client/src/main/kotlin/com/openlattice/directory/Auth0UserDirectoryService.kt
UnsungHero97: https://jira.openlattice.com/browse/LATTICE-2805

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
UnsungHero97: Updated.

katherinebernstein (approved):

api/src/main/kotlin/com/openlattice/search/Auth0UserSearchFields.kt
makosblade: Gotcha, yeah that makes sense

makosblade (approved):

PR #183, opened 2021-05-19
release > main

UnsungHero97:


makosblade (approved):

makosblade (commented):
actually we shouldn't merge main into this before merging it into main because now I don't think it represents what was acutally released

katherinebernstein (approved):

UnsungHero97:
> we shouldn't merge main into this before merging it into main because now I don't think it represents what was actually released

You're 100% right; it was an accident.

PR #189, opened 2021-05-27
CollaborationsApi naming

UnsungHero97:
The controller was not actually deleted, just renamed.

### DIFF
--- a/api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
+++ b/api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
@@ -1,97 +1,189 @@
 package com.openlattice.collaborations
 
+import com.openlattice.organizations.Organization
 import com.openlattice.organizations.OrganizationDatabase
+import com.openlattice.authorization.Permission
 import retrofit2.http.*
 import java.util.*
 
 interface CollaborationsApi {
 
     companion object {
-        // @formatter:off
-        const val  SERVICE = "/datastore"
-        const val  CONTROLLER = "/collaborations"
-        const val  BASE = SERVICE + CONTROLLER
-        // @formatter:on
+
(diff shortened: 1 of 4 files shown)


giuliacampana (approved):

api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
```diff
@@ -1,97 +1,189 @@
...
-    @POST(BASE + ID_PATH_PARAM + ORGANIZATIONS_PATH)
-    fun addOrganizationIdsToCollaboration(@Path(ID) id: UUID, @Body organizationIds: Set<UUID>)
+    /**
+     * Adds the given organization ids to the [Collaboration] object with the given collaboration id. The caller must
+     * have [Permission.OWNER] on the target [Collaboration] object.
+     *
+     * @param collaborationId [Collaboration] id
+     * @param organizationIds a set of [Organization] ids
+     */
+    @PATCH(BASE + COLLABORATION_ID_PATH + ORGANIZATIONS_PATH)
```
giuliacampana, line 83: just checking that the @POST to @PATCH change was intentional

giuliacampana (approved):

api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
UnsungHero97: Yes, it's an intentional change. The reason I'm switching it is because `PATCH` seems more appropriate (I think) since it's meant to be used for applying partial updates to an existing thing vs `POST` is more for creating new things.

geekbeast (dismissed):

giuliacampana (approved):

geekbeast (approved):

PR #190, opened 2021-05-30
Fix/neighbor search

geekbeast:
Switched to using allEntitySetIds. The paged requests aren't in great shape and it's not clear how the arguments are used.


geekbeast:
This picked up some changes from realtime linking controller.

UnsungHero97 (approved):

geekbeast:
Admin override with 1 approval.

PR #191, opened 2021-06-03
Feature/server side entity id generation

geekbeast:
Exposes an endpoint to make it easy for frontend applications to assign entity key ids, based off a subset of properties.


UnsungHero97 (approved):

geekbeast:
Overriding with 1 approval.

PR #201, opened 2021-06-10
Fix/cleanup datastore elasticsearch impls

geekbeast:
Should do thorough testing of this-- it is unclear why we were still using the non-Kotlin version.


datastore/src/main/java/com/openlattice/datastore/pods/DatastoreServicesPod.java
```diff
@@ -549,7 +550,7 @@ public AwsAuth0TokenProvider auth0TokenProvider() {
 
     @Bean
     public ConductorElasticsearchApi conductorElasticsearchApi() {
-        return new DatastoreElasticsearchImpl( datastoreConfiguration.getSearchConfiguration() );
+        return new DatastoreKotlinElasticsearchImpl(  datastoreConfiguration.getSearchConfiguration() );
```
UnsungHero97, line 553: Why not `ConductorElasticsearchImpl`?

UnsungHero97 (approved):

geekbeast:
Overriding and merging with 1 approval.

PR #208, opened 2021-06-24
Make compatible with gradle 7.

geekbeast:


geekbeast:
Admin override merge to enable gradle 7 builds

PR #209, opened 2021-06-24
Update repos.gradle

geekbeast:
Gradle doesn't like http anymore


geekbeast:
Overriding and merging for gradle 7 upgrade.

PR #218, opened 2021-07-16
add logging

anzioka:
PR:
Log error when PartitionManager.getDefaultPartitions(organizationId) returns null. 

This is an attempt to figure out why CollectionsManager.addTypeToEntityTypeCollectionTemplate(entityTypeCollectionId, collectionTemplateType) results in an error: `null cannot be cast to non-null type com.geekbeast.rhizome.hazelcast.DelegatedIntList`


conductor-client/src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
```diff
@@ -58,7 +58,12 @@ class PartitionManager @JvmOverloads constructor(
     }
 
     fun getDefaultPartitions(organizationId: UUID): List<Int> {
-        return organizations.executeOnKey(organizationId, OrganizationReadEntryProcessor { DelegatedIntList(it.partitions) }) as DelegatedIntList
+        val partitions = organizations.executeOnKey(organizationId, OrganizationReadEntryProcessor { DelegatedIntList(it.partitions) })
+        if (partitions == null) {
+            logger.error("ERROR: loading default partitions for organization {} returned null", organizationId);
```
UnsungHero97, line 63: `logger.error` will already print the `ERROR` prefix.

```suggestion
            logger.error("loading default partitions for organization {} returned null", organizationId);
```

UnsungHero97 (approved):

UnsungHero97 (approved):

conductor-client/src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
```diff
@@ -58,7 +58,10 @@ class PartitionManager @JvmOverloads constructor(
     }
 
     fun getDefaultPartitions(organizationId: UUID): List<Int> {
-        return organizations.executeOnKey(organizationId, OrganizationReadEntryProcessor { DelegatedIntList(it.partitions) }) as DelegatedIntList
+        val partitions = organizations.executeOnKey(organizationId, OrganizationReadEntryProcessor { DelegatedIntList(it.partitions) })
+        checkNotNull(partitions, {"default partitions for organization $organizationId should not be null"})
```
geekbeast, line 62: Doesn't kotlin allow checkNotNull(...) {...}? Minor style change as template is only evaluated on null either way.

geekbeast (approved):
Style comment, lgtm

conductor-client/src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
anzioka: ah, yes
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check-not-null.html

PR #226, opened 2021-08-24
WIP - Task/LATTICE-2887-docker Getting Docker up and running for the stack

rbachchhs:


Docker-README.md
```diff
@@ -0,0 +1,59 @@
...
+The fundamental structure is the same as our architecture although postgres and elasticsearch will not be containerized as they do not run particularly well inside of containers given their resource needs
+
+The usage process is as follows:
+
+- The docker-compose file (for better or worse) relies on having a locally built version of each service in their respective build/distributions/ folders
+build and dist our software packages.
+	- I prefer this for building:
+		- `gradle -q --build-cache --daemon --parallel build -x test -x spotbugsMain -x spotbugsTest -x checkstyleMain -x checkstyleTest -x javadoc -x distZip -x distTar`
+	- and this for creating dists:
+		- `gradle -q --build-cache --daemon --parallel :conductor:distTar :datastore:distTar :indexer:distTar -x test -x spotbugsMain -x spotbugsTest -x checkstyleMain -x checkstyleTest -x javadoc -x distZip`
```
geekbeast, line 14: isn't --parallel and --daemon the default now?

Docker-README.md
```diff
@@ -0,0 +1,59 @@
...
+
+- The IP address/host running postgres/elasticsearch must be manually specified in the configuration files as there is no operating-system agnostic placeholder that docker can interpret as 'the host running docker' (your local machine)
+
+- Services have health checks configured and exposed to docker's healthy host monitoring system so if it seems like something isn't starting up and it should be, make sure that all other container dependencies are starting up and becoming healthy
+- You may need to increase the compose healthcheck timeouts for ^
+
+
+Work still to do:
+
+- Add an auth0syncTask that creates test users instead of syncing down real users
```
geekbeast, line 49: I think this may already exist as LocalAuth0SyncTask instead of DefaultAuth0SyncTask

UnsungHero97: I also created a new tenant for testing and seeded it with some test users, so it would just be a matter of updating the config (I think).

PR #239, opened 2021-08-31
split getCollaborationsWithId from getCollaborations

solomontang:


solomontang:
see https://github.com/openlattice/openlattice/pull/241

PR #240, opened 2021-09-01
Modified RetrofitFactory urls to point to the Canadian server

rekahalmai:
Modified RetrofitFactory urls to point to the Canadian server

### DIFF
--- a/api/src/main/java/com/openlattice/client/RetrofitFactory.java
+++ b/api/src/main/java/com/openlattice/client/RetrofitFactory.java
@@ -34,12 +34,12 @@ import retrofit2.CallAdapter;
 import retrofit2.Retrofit;
 
 public final class RetrofitFactory {
-    private static final String BASE_URL                = "https://api.openlattice.com/";
-    private static final String CHILDCARE_URL           = "https://api.childcare.openlattice.com";
-    private static final String INTEGRATION_URL         = "https://integration.openlattice.com/";
+    private static final String BASE_URL                = "https://api.ca.openlattice.com/";
+    private static final String CHILDCARE_URL           = "https://api.childcare.ca.openlattice.com";
+    private static final String INTEGRATION_URL         = "https://integration.ca.openlattice.com/";
     private static final String LOCAL_BASE_URL          = "http://localhost:8080/";
-    private static final String STAGING_BASE_URL        = "https://api.staging.openlattice.com/";
-    private static final String STAGING_INTEGRATION_URL = "https://integration.staging.openlattice.com/";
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/openlattice/client/RetrofitFactory.java
```diff
@@ -34,12 +34,12 @@
...
-    private static final String CHILDCARE_URL           = "https://api.childcare.openlattice.com";
-    private static final String INTEGRATION_URL         = "https://integration.openlattice.com/";
+    private static final String BASE_URL                = "https://api.ca.openlattice.com/";
+    private static final String CHILDCARE_URL           = "https://api.childcare.ca.openlattice.com";
+    private static final String INTEGRATION_URL         = "https://integration.ca.openlattice.com/";
     private static final String LOCAL_BASE_URL          = "http://localhost:8080/";
-    private static final String STAGING_BASE_URL        = "https://api.staging.openlattice.com/";
-    private static final String STAGING_INTEGRATION_URL = "https://integration.staging.openlattice.com/";
+    private static final String STAGING_BASE_URL        = "https://api.staging.ca.openlattice.com/";
+    private static final String STAGING_INTEGRATION_URL = "https://integration.staging.ca.openlattice.com/";
```
UnsungHero97, line 42: Instead of overwriting the existing environments, you should add new ones specifically for Canada. For example:

```
private static final String BASE_URL = "https://api.openlattice.com/";
private static final String BASE_CA_URL = "https://api.ca.openlattice.com/";
```

Additionally, we don't need a copy of all the urls right now (there is no staging or childcare in the Canada stack), so I think `BASE_URL` and `INTEGRATION_URL` are the only ones you need.

UnsungHero97 (changes requested):

api/src/main/java/com/openlattice/client/RetrofitFactory.java
rekahalmai: @UnsungHero97 Done but just a quick question - these new constants are not in use, shouldn't we change anything else? I understood that this branch would be deployed on the Canadian server (and not merged with the main), that's why I modified directly the existing env constants. Let me know if I'm missing something :)

UnsungHero97: Hmmm, I don't think we've talked about deploying different branches for Canada. Additionally, shuttle doesn't get "deployed" ... it just needs to pull in this updated code so that when you run shuttle, it's aware of the new environments and can talk to the Canada stack.

solomontang (approved):

kengie (approved):

geekbeast (approved):

PR #241, opened 2021-09-02
Feature/lattice 2892 collaboration api enhancements

solomontang:
- `getCollaborations` has been reworked to get Collaboration objects (READ) for each collaboration id provided
- `getAllCollaborations` gets all Collaboration objects the caller has READ permissions on (old getCollaborations behavior)
- `getCollaborationsWithDataSets` has been reworked to return Collaboration objects, instead of just their ids

### DIFF
--- a/api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
+++ b/api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
@@ -40,7 +40,7 @@ interface CollaborationsApi {
     /**
      * Gets [Collaboration] objects the caller has [Permission.READ] on with the provided ids.
      *
-     * @return a list of [Collaboration] objects
+     * @return a Map<K, V> where K is a collaboration id and V a [Collaboration] object
      */
     @GET(BASE)
     fun getCollaborations(@Query(ID_PARAM) ids: Set<UUID>): Map<UUID, Collaboration>
@@ -192,7 +192,7 @@ interface CollaborationsApi {
      * the caller has [Permission.READ] on.
      *
      * @param dataSetIds a set of data set ids
-     * @return Map<K, V> where K is a data set id and V is a list of [Collaboration] ids
+     * @return Map<K, V> where K is a data set id and V is a list of [Collaboration] objects
      */
     @POST(BASE + DATA_SETS_PATH)
     fun getCollaborationsWithDataSets(@Body dataSetIds: Set<UUID>): Map<UUID, List<Collaboration>>
--- a/api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
+++ b/api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
@@ -24,7 +24,7 @@ interface CollaborationsApi {
         const val COLLABORATION_ID_PATH = "/{$COLLABORATION_ID_PARAM}"
         const val DATA_SET_ID_PARAM = "dataSetId"
         const val DATA_SET_ID_PATH = "/{$DATA_SET_ID_PARAM}"
-        const val ID_PARAM = "id"
+        const val IDS_PARAM = "ids"
         const val ORGANIZATION_ID_PARAM = "organizationId"
         const val ORGANIZATION_ID_PATH = "/{$ORGANIZATION_ID_PARAM}"
     }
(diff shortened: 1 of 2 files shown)


datastore/src/main/kotlin/com/openlattice/collaborations/controllers/CollaborationsController.kt
```diff
@@ -41,14 +43,24 @@ class CollaborationsController : AuthorizingComponent, CollaborationsApi {
...
         val authorizedCollaborationIds = getAllAuthorizedCollaborationIds()
         return collaborationService.getCollaborations(authorizedCollaborationIds).values
     }
 
+    @Timed
+    @GetMapping(
+            value = ["", "/"],
+            produces = [MediaType.APPLICATION_JSON_VALUE]
+    )
+    override fun getCollaborations(@RequestParam( ID_PARAM ) ids: Set<UUID>): Map<UUID, Collaboration> {
```
geekbeast, line 59: Do you want this in as the body or in the URL? That URL will get long real fast.

api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
```diff
@@ -23,6 +24,7 @@ interface CollaborationsApi {
         const val COLLABORATION_ID_PATH = "/{$COLLABORATION_ID_PARAM}"
         const val DATA_SET_ID_PARAM = "dataSetId"
         const val DATA_SET_ID_PATH = "/{$DATA_SET_ID_PARAM}"
+        const val ID_PARAM = "id"
```
UnsungHero97, line 27:
```suggestion
        const val IDS_PARAM = "ids"
```

api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
```diff
@@ -32,8 +34,16 @@ interface CollaborationsApi {
...
+    fun getAllCollaborations(): Iterable<Collaboration>
+
+    /**
+     * Gets [Collaboration] objects the caller has [Permission.READ] on with the provided ids.
+     *
+     * @return a Map<K, V> where K is a collaboration id and V a [Collaboration] object
+     */
     @GET(BASE)
-    fun getCollaborations(): Iterable<Collaboration>
+    fun getCollaborations(@Query(ID_PARAM) ids: Set<UUID>): Map<UUID, Collaboration>
```
UnsungHero97, line 46:
```suggestion
    fun getCollaborations(@Query(IDS_PARAM) ids: Set<UUID>): Map<UUID, Collaboration>
```

datastore/src/main/kotlin/com/openlattice/collaborations/controllers/CollaborationsController.kt
UnsungHero97:
```suggestion
    override fun getCollaborations(@RequestParam( IDS_PARAM ) ids: Set<UUID>): Map<UUID, Collaboration> {
```

UnsungHero97 (changes requested):

api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
solomontang: the param should be id. if multiple ids are passed, they're each assigned their own in repeat
`?id=1234&id=4567`

datastore/src/main/kotlin/com/openlattice/collaborations/controllers/CollaborationsController.kt
solomontang: using GET with a body, or implementing POST with body with the intention of using it as a GET both deviate from what is expect from REST.

length of url isn't really a concern. if we really run into character limitations, it should be handled by limiting API calls to a certain number of ids per request.

api/src/main/kotlin/com/openlattice/collaborations/CollaborationsApi.kt
UnsungHero97: We discussed this in slack, agreeing to keep it as `id`.

UnsungHero97 (dismissed):

datastore/src/main/kotlin/com/openlattice/collaborations/controllers/CollaborationsController.kt
geekbeast: The comment about GET makes sense, though I am not as devoted to REST philosophy when (ab)using it for RPC. You'll have to deal with paging eventually so I think this is fine for now.

geekbeast: Also there is a trade off here for debugging/testing this API, but I know where those tickets will go :)

UnsungHero97 (approved):

geekbeast (approved):

PR #249, opened 2021-09-17
Feature/unversioned property updates

geekbeast:
This should significantly reduce the vacuum pressure on our stack from integrations by not overwriting existing properties.


UnsungHero97 (changes requested):

api/src/main/kotlin/com/openlattice/data/PropertyUpdateType.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+enum class PropertyUpdateType {
+    /**
+     * Indicates that an additional version should be recorded in the versions array.
+     */
+    Versioned,
+    /**
+     * Indicates that an additional version should not be recorded in the versions array.
+     */
+    Unversioned
```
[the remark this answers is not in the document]
geekbeast: There's a lot of discussion on this and kotlin style guide says both are acceptable (https://kotlinlang.org/docs/coding-conventions.html#property-names). Uppercase seems to cause some issues with swagger codegen (https://github.com/swagger-api/swagger-codegen/issues/7466), but I don't think its relevant hear so I'm happy to adopt upper case.

datastore/src/main/kotlin/com/openlattice/datastore/data/controllers/DataIntegrationController.kt
```diff
@@ -47,14 +46,21 @@ class DataIntegrationController : DataIntegrationApi, AuthorizingComponent {
...
     }
 
     @Timed
     @PostMapping("/" + DataIntegrationApi.S3)
     fun generatePresignedUrls(
-        @RequestBody data: List<S3EntityData>
+            @RequestBody data: List<S3EntityData>,
+            @RequestParam(
+                    name = PROPERTY_UPDATE_TYPE,
+                    defaultValue = "Versioned"
```
anzioka, line 62: `PropertyUpdateType.Versioned` rather than hard-coded string

UnsungHero97: I don't think that's possible.

Screen Shot 2021-09-29 at 5 54 53 PM

UnsungHero97 (approved):

anzioka (approved):

PR #251, opened 2021-09-21
Task/lattice 2876 updated migration

njhlai:


conductor-client/src/main/java/com/openlattice/postgres/PostgresTable.java
```diff
@@ -386,7 +386,17 @@
                             ID_MAP,
                             VERSION )
                     .distributionColumn( LINKING_ID );
-
+    // needed for migration
+    public static final  PostgresTableDefinition LEGACY_PERMISSIONS        =
```
geekbeast, line 390: Understand why we have this, but should also make sure we have SQL to create this iron'd out (should be a simple inner join filtering by securable object types).

geekbeast (approved):

conductor-client/src/main/java/com/openlattice/postgres/PostgresTable.java
njhlai: Yup, I have the SQL detailed in the migration ticket itself.

UnsungHero97 (approved):

PR #252, opened 2021-09-24
add entityTypeId to DataSet

bergieboy:


anzioka (approved):

geekbeast (approved):

UnsungHero97 (approved):
This should be enough to unblock the frontend since the data set page calls `DataSetMetadataApi.getDataSetMetadata`. However, `entityTypeId` will not be included when calling `SearchApi.searchDataSetMetadata` until we reindex the data in elasticsearch.

solomontang (approved):

PR #253, opened 2021-09-24
Lattice 2709 org data warehouse support

and-carter:
Add CRUD endpoints to AdminController for adding system wide data warehouses for Redshift/Snowflake/Aurora/etc

Need to fix issue where GET only reads from memory. Warehouses in postrges not being fetched on GET request.

Done
- Add CRUD endpoints to AdminAPI and AdminController
- Add WarehouseService
- Create JdbcConnectionParameters object
- Add Hazelcast support for Warehouses
- Add Postgres table for Warehouses
- Add WarehouseMapstore


api/src/main/kotlin/com/openlattice/organizations/JdbcConnectionParameters.kt
```diff
@@ -0,0 +1,56 @@
...
+constructor(
+    @JsonProperty(SerializationConstants.ID_FIELD) val _id: UUID = UUID.randomUUID(),
+    @JsonProperty(SerializationConstants.TITLE_FIELD) val _title: String,
+    @JsonProperty(SerializationConstants.URL) val url: String,
+    @JsonProperty(SerializationConstants.DRIVER) val driver: String,
+    @JsonProperty(SerializationConstants.DATABASE) val database: String = "",
+    @JsonProperty(SerializationConstants.USERNAME) val username: String = "",
+    @JsonProperty(SerializationConstants.PASSWORD) val password: String = "",
+    @JsonProperty(SerializationConstants.PROPERTIES_FIELD) val properties: Properties = Properties(),
+    @JsonProperty(SerializationConstants.DESCRIPTION_FIELD) val description: Optional<String> = Optional.empty()
```
UnsungHero97, line 44: Why did you go with `Optional<String>` instead of doing what you did above?

api/src/main/kotlin/com/openlattice/organizations/JdbcConnectionParameters.kt
```diff
@@ -0,0 +1,56 @@
...
+ * @author Matthew Tamayo-Rios &lt;matthew@openlattice.com&gt;
+ */
+
+@JsonFilter("")
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+class JdbcConnectionParameters
+/**
+ * Creates an instance of a jdbc connection. All class properties must be serializable or marked as transient.
+ *
+ * @param _title Title of the connection.
```
UnsungHero97, line 24: Missing `_id`.

api/src/main/kotlin/com/openlattice/admin/AdminApi.kt
```diff
@@ -36,11 +38,14 @@ const val ID = "id"
 const val ID_PATH = "/{$ID}"
 const val NAME = "name"
 const val NAME_PATH = "/{$NAME}"
+const val WAREHOUSE_ID_PARAM = "WarehouseId"
```
UnsungHero97, line 41: The convention is camel case.

```suggestion
const val WAREHOUSE_ID_PARAM = "warehouseId"
```

api/src/main/kotlin/com/openlattice/admin/AdminApi.kt
```diff
@@ -36,11 +38,14 @@ const val ID = "id"
...
+const val WAREHOUSE_ID_PATH = "/{${WAREHOUSE_ID_PARAM}}"
 
 const val JOBS = "/jobs"
 
-interface AdminApi {
+const val WAREHOUSES = "/warehouses"
```
UnsungHero97, line 46: Our conventions are not consistent here, but what I've been doing is ... if it's part of the path, use `_PATH`, and if it's a param (path or otherwise), use `_PARAM`. So ...

```suggestion
const val WAREHOUSES_PATH = "/warehouses"
```

api/src/main/kotlin/com/openlattice/organizations/JdbcConnectionParameters.kt
```diff
@@ -0,0 +1,56 @@
...
+ * @param username Username used to log onto the database.
+ * @param password Password used to log onto the database.
+ * @param properties Object containing properties of the connection.
+ * @param description Description of the database.
+ *
+ */
+@JvmOverloads
+constructor(
+    @JsonProperty(SerializationConstants.ID_FIELD) val _id: UUID = UUID.randomUUID(),
+    @JsonProperty(SerializationConstants.TITLE_FIELD) val _title: String,
```
UnsungHero97, line 37: Why the `_`?

api/src/main/kotlin/com/openlattice/admin/AdminApi.kt
```diff
@@ -101,4 +106,51 @@ interface AdminApi {
...
+    /**
+     *  Gets JDBC connection details for a specified Warehouse
+     *
+     *   @param WarehouseId the ID for the warehouse to get details about.
+     *   @return The JDBC connection details that define the warehouse.
+     *   - Fields: Title, Url, Driver, Database, Username, Password, Properties, Description
+     *
+     */
+    @GET(BASE + WAREHOUSES + WAREHOUSE_ID_PATH)
+    fun getWarehouse(@Path(WAREHOUSE_ID_PARAM) WarehouseId: UUID): JdbcConnectionParameters
```
UnsungHero97, line 128:
```suggestion
    fun getWarehouse(@Path(WAREHOUSE_ID_PARAM) warehouseId: UUID): JdbcConnectionParameters
```

api/src/main/kotlin/com/openlattice/admin/AdminApi.kt
```diff
@@ -101,4 +106,51 @@ interface AdminApi {
...
+    @POST(BASE + WAREHOUSES)
+    fun createWarehouse(@Body jdbc: JdbcConnectionParameters): UUID
+
+    /**
+     *  Deletes a specified warehouse from the known warehouse connections.
+     *
+     *  @param WarehouseID The ID of the warehouse to be deleted.
+     */
+    @DELETE(BASE + WAREHOUSES + WAREHOUSE_ID_PATH)
+    fun deleteWarehouse(@Path(WAREHOUSE_ID_PARAM) WarehouseId: UUID)
```
UnsungHero97, line 147:
```suggestion
    fun deleteWarehouse(@Path(WAREHOUSE_ID_PARAM) warehouseId: UUID)
```

api/src/main/kotlin/com/openlattice/admin/AdminApi.kt
```diff
@@ -101,4 +106,51 @@ interface AdminApi {
...
+
+    /**
+     * Update the connection details for a warehouse JDBC connection.
+     *
+     *  @param jdbc JdbcConnectionParameters to replace existing parameters. Jdbc.id indicates the Warehouse to be updated.
+     */
+    @PATCH(BASE + WAREHOUSES)
+    fun updateWarehouse(@Body jdbc: JdbcConnectionParameters): Int
```
UnsungHero97, line 155: If I only want to update a single param (or just a few), do I still need to pass the entire object with every param filled in? If so, that's not ideal.

conductor-client/src/main/java/com/openlattice/hazelcast/pods/MapstoresPod.java
```diff
@@ -314,6 +315,11 @@ public CollaborationMapstore collaborationMapstore() {
         return new CollaborationMapstore( hikariDataSource );
     }
 
+    @Bean
+    public WarehousesMapstore warehouseMapstore() {
```
UnsungHero97, line 319:
```suggestion
    public WarehousesMapstore warehousesMapstore() {
```

conductor-client/src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -75,6 +75,7 @@
 import com.openlattice.organization.ExternalTable;
 import com.openlattice.organization.OrganizationEntitySetFlag;
 import com.openlattice.organization.roles.Role;
+import com.openlattice.organizations.JdbcConnectionParameters;
```
UnsungHero97, line 78: It looks like the only changes to this file are imports. If they are not used anywhere, please remove.

conductor-client/src/main/kotlin/com/openlattice/organizations/WarehouseService.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.openlattice.organizations.JdbcConnectionParameters
+import org.springframework.stereotype.Service
+import java.util.UUID
+
+/**
+ * @author Andrew Carter andrew@openlattice.com
+ */
+
+@Service
+class WarehouseService(
```
UnsungHero97, line 19: I think it would be more consistent if it was plural ... warehouses mapstore, warehouses servce, etc.

```suggestion
class WarehousesService(
```

conductor-client/src/main/kotlin/com/openlattice/organizations/WarehouseService.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    fun deleteWarehouse(id: UUID) {
+        ensureValidWarehouseId(id)
+        warehouses.delete(id)
+    }
+
+    fun updateWarehouse(jdbc: JdbcConnectionParameters): Int {
+        ensureValidWarehouseId(jdbc._id)
+        warehouses.replace(jdbc._id, jdbc)
+        return 1
```
UnsungHero97, line 49: Is there a reason why you want to return an int here? I don't see the value. If the request returns with HTTP 200, then I know the update was successful.

UnsungHero97 (changes requested):

api/src/main/kotlin/com/openlattice/organizations/JdbcConnectionParameters.kt
and-carter: No reasoning other than that's what I saw Matthew do in his old version of this code. 
https://github.com/openlattice/openlattice/pull/217/files#diff-afa5ae07f36c1dcd234152675bb6b602ec27c350a000337925ad28b4e1d1dff5R37

api/src/main/kotlin/com/openlattice/organizations/JdbcConnectionParameters.kt
and-carter: Without `_` JdbcConnectionParameters fails to compile because of an `Accidental override` error. 

I believe this happens because the Kotlin compiler tries to generate a getter for `val id` declared in the class primary constructor but the base class, `AbstractSecurableObject`, already has one. 

Not sure if there is a better way to avoid this, but I saw it was done this way in the Collaborations class so I followed it as convention 
https://github.com/openlattice/openlattice/blob/f89b8d18315fe3186a378ee10b26a80ce2319cf8/api/src/main/kotlin/com/openlattice/collaborations/Collaboration.kt#L12

One error output I saved for reference:
`Accidental override: The following declarations have the same JVM signature (getId()Ljava/util/UUID;`

api/src/main/kotlin/com/openlattice/admin/AdminApi.kt
and-carter: Yes, that's how it's implemented now. Will add endpoints for individual params.

conductor-client/src/main/kotlin/com/openlattice/organizations/WarehouseService.kt
and-carter: Ah yes that is redundant. I'll take that out.

UnsungHero97 (approved):

PR #254, opened 2021-09-27
implement endpoint to delete entities and specified neighbors

anzioka:


api/src/main/java/com/openlattice/data/DataApi.java
```diff
@@ -194,22 +187,24 @@ Integer replaceAssociationData(
...
-     * @param deleteType The delete type to perform (soft or hard delete).
+     * @param entitySetId          The id of the EntitySet to delete from.
+     * @param filter               EntityNeighboursFilter containing which ids of entities to delete and entity set ids of neighbours
+     *                             to delete from.
+     * @param deleteType           The delete type to perform (soft or hard delete).
+     * @param blockUntilCompletion If set to true, controller will wait for deletion to complete before returning to caller
      */
-    @POST( BASE + "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + NEIGHBORS )
-    Long deleteEntitiesAndNeighbors(
+    @HTTP( method = "DELETE", path = BASE + "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + NEIGHBORS, hasBody = true )
```
UnsungHero97, line 196: `@DELETE`

anzioka: Endpoint has a request body. 
```
java.lang.IllegalArgumentException: Non-body HTTP method cannot contain @Body.
    for method DataApi.deleteEntitiesAndNeighbors
```

conductor-client/src/main/kotlin/com/openlattice/data/jobs/EntitiesAndNeighborsDeletionJob.kt
```diff
@@ -0,0 +1,134 @@
...
+        if (entitySetId == null) {
+            logger.info("No more delete jobs to submit for processing")
+            if (allDeleteJobsFinished()) {
+                hasWorkRemaining = false
+                status = JobStatus.FINISHED
+                updateJobStatus()
+                publishJobState()
+                logger.info("finished deleting entities and neighbors of entity set ${state.entitySetId}")
+            }
+            Thread.sleep(STATUS_CHECK_INTERVAL)
```
geekbeast, line 60: Polling is handled at a higher layer and this indicates something might be wrong with the design.

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -907,35 +903,65 @@ public Integer deleteEntityProperties(
...
+        Set<UUID> allEntitySetIds = Sets.union( srcEntitySetIds, dstEntitySetIds );
+        allEntitySetIds = Sets.union( allEntitySetIds, Set.of( entitySetId ) );
+        
+        ensureEntitySetsCanBeWritten( allEntitySetIds );
+
+        Map<UUID, Set<UUID>> entitySetIdEntityKeyIds = Maps.newHashMap();
+        entitySetIdEntityKeyIds.put( entitySetId, filter.getEntityKeyIds() );
+
+        // verify permission to delete
+        entitySetIdEntityKeyIds.forEach( ( key, val ) -> {
```
geekbeast, line 932: This seems like both a bad idea and inefficient.

geekbeast: So deeper explanation his will hit postgres with expensive queries several times. The problem is that none of the functions in the chain allow for batch parameters and the limiting step (an edge query to determine which entity sets are potential neighbor for auth checks) is expensive. Without digging into the query plan generated by postgres (most likely a sequential scan) my guess is that you'd get the same speed by rolling all of them up. In addition-- each call invokes an unnecessary query against postgres( Graph.checkForUnauthorized ). This isn't necessary because you already access checked all potential entity sets which would include any edges checked by checkForUnauthorized. We should do one or the other. Graph.checkForUnauthorized seems like the better approach since it limits edges to the edges affected by the delete and not all potential neighbor entity sets.

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -1223,6 +1249,27 @@ private int blockOnDeletionJobGettingNumUpdates( UUID deletionJobId, boolean blo
         return 0;
     }
 
+    // Block until JobStatus is FINISHED or CANCELED
+    public void waitForDeleteJobToTerminate( UUID deletionJobId ) {
```
geekbeast, line 1253: This should be in job service. The most straight forward way is to introduce a job service topic and make the job service subscribe to it. Whenever a job is completed it fires a JobCompletedEvent that contains its job id and causes the service to release the lock held in a *local* Map<UUID, CountDownLatch>. Once the latch successfully counts down, remove the entry from the local map if it is still there.

geekbeast: See ReliableTopic in Hazelcast.

conductor-client/src/main/kotlin/com/openlattice/data/jobs/EntitiesAndNeighborsDeletionJob.kt
```diff
@@ -0,0 +1,134 @@
...
+        if (entityKeyIds.isEmpty()) {
+            logger.warn("skipping entity set $entitySetId since there are no entities to delete")
+            cleanUpBatch(entitySetId)
+            publishJobState()
+            return
+        }
+
+        val jobService = lateInitProvider.dataGraphService.getJobService()
+
+        val jobId = jobService.submitJob(DataDeletionJob(DataDeletionJobState(
```
geekbeast, line 80: Clever, but this means DataDeletionJob should be cleaned up or all sorts of fun bad things can happen.

geekbeast (changes requested):
Needs some more work. The approach of creating data deletion jobs is interesting, but won't work without fixing the issues with the current data deletion jobs. It maybe easier to introduce a delete neighbors flag to the existing data deletion job.  Also this doesn't actually seem to batch off of postgres and instead seems to load all entity key ids into RAM and then pages through them on an entity set basis. For Chronicle this will cause resource usages issues have to page in postgres.

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
anzioka: I added this primarily for testing. There is a rehearsal test that verifies that the data has actually been deleted, hence needs to wait until completion.

datastore/src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -320,7 +320,7 @@ constructor(
         )
         timer.reset().start()
 
-        deletionManager.authCheckForEntitySetAndItsNeighbors(entitySetId, deleteType, Principals.getCurrentPrincipals())
+        deletionManager.authCheckForEntitySetsAndNeighbors(setOf(entitySetId), DeleteType.Hard, Principals.getCurrentPrincipals())
```
UnsungHero97, line 323: We specifically made the change to remove `DeleteType.Hard` as being the default. This should remain as `deleteType`; please revert.

UnsungHero97 (changes requested):

conductor-client/src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -977,6 +977,38 @@ class Graph(
...
+        val query = """
+            SELECT DISTINCT ${EDGE_ENTITY_SET_ID.name} FROM ${E.name}
+            WHERE ( ${DST_ENTITY_SET_ID.name} = ANY(?) $dstEntityKeyIdFilter )
+            OR ( ${SRC_ENTITY_SET_ID.name} = ANY(?) $srcEntityKeyIdFilter )
+        """.trimIndent()
+
+        val reader = dataSourceResolver.getDefaultDataSource()
+        return BasePostgresIterable(PreparedStatementHolderSupplier(reader, query) { ps ->
+            val entitySetIdsArr = PostgresArrays.createUuidArray(ps.connection, entitySetIds)
+            var index = 0
```
UnsungHero97, line 996: No change required, just a thought on structure. I think it's more straightforward to initialize index to 1 and do `index++` (or maybe even skip it altogether).

conductor-client/src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -116,41 +112,51 @@ class DataDeletionService(
     /* Authorization checks */
 
     @Timed
-    override fun authCheckForEntitySetAndItsNeighbors(entitySetId: UUID, deleteType: DeleteType, principals: Set<Principal>, entityKeyIds: Set<UUID>?) {
-        val isAssociationEntitySet = entitySetManager.getEntitySet(entitySetId)!!.flags.contains(EntitySetFlag.ASSOCIATION)
+    override fun authCheckForEntitySetsAndNeighbors(entitySetIds: Set<UUID>, deleteType: DeleteType, principals: Set<Principal>, entityKeyIds: Set<UUID>?) {
+        val srcDstEntitySets = entitySetIds.filter {
+            !entitySetManager.getEntitySet(it)!!.flags.contains(EntitySetFlag.ASSOCIATION)
+        }.toSet()
+        val neighborEdgeEntitySets = if (srcDstEntitySets.isEmpty()) mutableSetOf() else graphService.getNeighborEdgeEntitySets(srcDstEntitySets, entityKeyIds)
```
UnsungHero97, line 119: One difference between this and the old code is that `neighborEdgeEntitySets` has not gone through the permissions filter yet. See `getAuthorizedNeighborEntitySetsQuery`, though that query doesn't check for a specific permission, which seems bad.

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -907,35 +853,57 @@ public Integer deleteEntityProperties(
...
+        Map<UUID, Set<UUID>> entitySetIdEntityKeyIds = Maps.newHashMap();
+        entitySetIdEntityKeyIds.put( entitySetId, filter.getEntityKeyIds() );
+
+        // verify permission to delete
+        deletionManager.authCheckForEntitySetsAndNeighbors(
+                allEntitySetIds,
+                deleteType,
+                Principals.getCurrentPrincipals(), filter.getEntityKeyIds() );
+
+        UUID deletionJobId = deletionManager.clearOrDeleteEntitiesAndNeighbors(
```
UnsungHero97, line 887: As far as I can tell, there has been no permissions check on the entity sets themselves up to this point. `ensureEntitySetsCanBeWritten` doesn't actually check permissions and `authCheckForEntitySetsAndNeighbors` only checks permissions on properties.

I think you  need to call `checkPermissionsOnEntitySetIds` instead of `ensureEntitySetsCanBeWritten`. However, this is not enough because it doesn't consider the edge entity sets.

UnsungHero97 (changes requested):

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -1195,6 +1163,10 @@ private void ensureEntitySetsCanBeWritten( Set<UUID> entitySetIds ) {
         }
     }
 
+    private void ensureEntitySetsAreNotEdges( Set<UUID> entitySetIds ) {
+
+    }
+
```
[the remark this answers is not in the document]
anzioka: Nope. Removed

conductor-client/src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
anzioka: Doesn't authorizationManager.accessChecksForPrincipal accomplish this? The AclKey includes both the entity set id and property type ids

datastore/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
anzioka: https://github.com/openlattice/openlattice/pull/254#discussion_r740719901

conductor-client/src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
UnsungHero97: In my opinion, no. It's a different check. It's checking property permissions, not entity set permissions. That being said, I'm not totally sure if it's necessary since it's strictly a delete on entities.

@geekbeast do you think it is necessary to check entity set permissions in addition to property permissions?

conductor-client/src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -113,9 +114,18 @@ class DataDeletionService(
...
+
+        val srcDstEntitySets = mutableSetOf<UUID>()
+        entitySetIds.forEach { esid ->
+            entitySetManager.getEntitySet(esid)?.let {
+                if (!it.flags.contains(EntitySetFlag.ASSOCIATION)) {
+                    srcDstEntitySets.add(esid)
+                }
+            } ?: run {
+                throw IllegalArgumentException("Unable to perform delete on $entitySetIds because $esid does not exist")
+            }
```
UnsungHero97, line 126: It's not guaranteed that this function will be called only during a delete.

UnsungHero97 (changes requested):

conductor-client/src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
anzioka: What do you mean by that? The function accepts a delete type as a parameter, and checks whether caller has correct permissions on affected entity sets and their property types based on the delete type.

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJob.kt
```diff
@@ -106,6 +92,13 @@ class DataDeletionJob(
             return
         }
 
+        // batch delete neighbors first. Only delete from state.entitySetId if there are no more neighbors to delete
+        val neighborEntityDataKeys = getBatchOfNeighborEntityDataKeys(entityDataKeys.map { it.entityKeyId }.toSet())
+        if (neighborEntityDataKeys.isNotEmpty()) {
+            deleteNeighborEntitiesAndEdges(neighborEntityDataKeys)
+            return
+        }
```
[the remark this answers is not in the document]
anzioka: The result of getBatchOfEntityDataKeys() is deleted after neighbors are deleted.

anzioka:
Awaiting https://github.com/openlattice/openlattice/pull/276 to be merged first

api/src/main/java/com/openlattice/data/DataApi.java
```diff
@@ -195,12 +196,12 @@ Integer replaceAssociationData(
...
-     * to delete from.
-     * @param deleteType The delete type to perform (soft or hard delete).
+     * @param filter      EntityNeighboursFilter containing which ids of entities to delete and entity set ids of neighbours
+     *                    to delete from.
+     * @param deleteType  The delete type to perform (soft or hard delete).
      */
-    @POST( BASE + "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + NEIGHBORS )
-    Long deleteEntitiesAndNeighbors(
+    @HTTP( method = "DELETE", path = BASE + "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + NEIGHBORS, hasBody = true )
+    UUID deleteEntitiesAndNeighbors(
```
njhlai, line 204: Need to update `openlattice.yaml`, specifically line 2149 in https://github.com/openlattice/openlattice/blob/6754c2a8899a8e619f09e24c4805836db23fe427/api/openlattice.yaml#L2148-L2151

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJob.kt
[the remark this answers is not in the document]
njhlai: Logic looks ok to me, but is the `return` necessary?

anzioka:
https://github.com/openlattice/openlattice/pull/254#discussion_r756510921
@njhlai `return` is necessary because deleteNeighborEntitiesAndEdges(neighborEntityDataKeys) deletes a batch of entityDataKeys and we want to make sure that _all_ neighbor entities are deleted before deleting from the main entity set. The next invocation of processNextBatch will delete the next batch of neighbor entity data keys.

conductor-client/src/main/kotlin/com/openlattice/data/jobs/DataDeletionJob.kt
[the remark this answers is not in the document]
and-carter: Looks good to me.

- I believe `return` is necessary so that when `processNextBatch()` is called, only a single batch of neighbors is deleted per function call as long as neighbors remain. While neighbors still remain we do not want to delete entities. Once all neighbors have been deleted, return is skipped as we now are okay to begin deleting all entities.
- And 104-110 are used to exit the function if there is no work to be done.

and-carter (approved):

geekbeast (approved):
Ship it if it works.

UnsungHero97 (approved):

PR #275, opened 2021-11-09
Use resolved data source for updating IDS version.

geekbeast:


conductor-client/src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -431,7 +431,7 @@ class PostgresEntityDataQueryService(
 
             //Make data visible by marking new version in ids table.
 
-            val updatedEntities = dataSourceResolver.getDataSource(DEFAULT_DATASOURCE).connection.use { idsConnection ->
+            val updatedEntities = hds.connection.use { idsConnection ->
```
UnsungHero97, line 434: You don't need to do `hds.connection` here because you can reuse the existing `connection` object, right? That way, you avoid opening a new connection to postgres.

conductor-client/src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -885,7 +885,7 @@ class PostgresEntityDataQueryService(
     ): BasePostgresIterable<UUID> {
         val partitions = partitionManager.getEntitySetPartitions(entitySetId)
         val hds =
-            dataSourceResolver.getDataSource(DEFAULT_DATASOURCE) //This query hits the ids table which is still centralized.
+            dataSourceResolver.resolve(entitySetId) //This query hits the ids table which is still centralized.
```
UnsungHero97, line 888: The comment is no longer necessary.

geekbeast, line 888:
```suggestion
            dataSourceResolver.resolve(entitySetId)
```

conductor-client/src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
geekbeast: Good catch! Fixed.

UnsungHero97 (approved):

giuliacampana (approved):

solomontang (approved):

PR #276, opened 2021-11-12
Remove partition references.

geekbeast:
This removes the partition column, partition manager, and re-partitioning job+api.

No changes to other functionality


conductor-client/src/main/kotlin/com/openlattice/data/storage/redshift/RedshiftDataTables.kt
```diff
@@ -0,0 +1,221 @@
...
+                        }
+                    }
+                }
+        )
+
+        @JvmStatic
+        fun buildDataTableDefinition(): PostgresTableDefinition {
+            val columns = dataTableColumns.toTypedArray()
+
+            val tableDefinition = CitusDistributedTableDefinition("data")
```
UnsungHero97, line 92: Since we're no longer using Citus, can't you use `PostgresTableDefinition` here instead (and remove `.distributionColumn(PARTITION)` below)?

conductor-client/src/main/kotlin/com/openlattice/data/storage/redshift/RedshiftDataTables.kt
```diff
@@ -0,0 +1,221 @@
...
+                    .addColumns(*columns)
+                    .primaryKey(ENTITY_SET_ID_STRING, ID_STRING, PROPERTY_TYPE_ID_STRING, HASH_STRING)
+                    .distributionColumn(PARTITION)
```
UnsungHero97, line 95: Why?

UnsungHero97 (dismissed):

conductor-client/src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -356,28 +330,16 @@ class PostgresLinkingQueryService(
             val linkingIdsArray = PostgresArrays.createUuidArray(ps.connection, linkingIds)
             /* Note: this inclusion may or may not speed up the function, depending how many partitions are
                covered by all the normal entity sets requested */
```
njhlai, line 332:
```suggestion
```
Comment no longer relevant

njhlai (commented):
I think mechanic was accidentally rebumped to the non-existent commit and reversing Hristo's fix, causing the current build error. It should be dfaac5 instead of 06d23a.

njhlai (commented):
I think `GetPartitionsFromEntitySetEntryProcessor.kt` (https://github.com/openlattice/openlattice/blob/main/conductor-client/src/main/kotlin/com/openlattice/edm/processors/GetPartitionsFromEntitySetEntryProcessor.kt) can also be deleted.

PR #291, opened 2021-12-15
Task/lattice 2973 linker memory

njhlai:
`SocratesMatcher` currently initialise with a copy of the model every time a new thread is started. These (copies of the) model actually live off heap according to dl4j’s documentation, and thus are not managed by the JVM’s gc. This can cause memory issues. 

Instead, wrap the model with `ParallelInference` and let dl4j manage the multi-threading.

EDIT: ```LINKER_OFFHEAP``` should be no more than half of ```LINKER_XMX```. Based on what I see on prod before, maybe something like ```LINKER_XMX=-Xmx50g```, ```LINKER_OFFHEAP=-Dorg.bytedeco.javacpp.maxbytes=20g``` should suffice.

### DIFF
--- a/linker/src/main/kotlin/com/openlattice/linking/matching/SocratesMatcher.kt
+++ b/linker/src/main/kotlin/com/openlattice/linking/matching/SocratesMatcher.kt
@@ -52,7 +52,7 @@ class SocratesMatcher(
         private val linkingFeedbackService: PostgresLinkingFeedbackService
 ) : Matcher {
 
-    private var localModel = ParallelInference.Builder(model)
+    private val localModel = ParallelInference.Builder(model)
                         .inferenceMode(InferenceMode.SEQUENTIAL)
                         // .batchLimit(32)
                         .workers(Runtime.getRuntime().availableProcessors())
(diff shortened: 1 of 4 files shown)


njhlai:
In production, `LINKER_OFFHEAP` needs to be defined, and `LINKER_XMX` and `LINKER_OFFHEAP` needs to be tuned as well.

linker/src/main/kotlin/com/openlattice/linking/matching/SocratesMatcher.kt
```diff
@@ -50,10 +52,14 @@ class SocratesMatcher(
         private val linkingFeedbackService: PostgresLinkingFeedbackService
 ) : Matcher {
 
-    private var localModel = ThreadLocal.withInitial { model.clone() }
+    private val localModel = ParallelInference.Builder(model)
+                        .inferenceMode(InferenceMode.SEQUENTIAL)
+                        // .batchLimit(32)
```
njhlai, line 57: Need decision whether to do this sequentially or in batches.

geekbeast: That doesn't sound very parallel.

njhlai: Batching makes the most sense here I think, just want to see if anyone has strong opinions/if I miss anything.

geekbeast (approved):

UnsungHero97 (approved):

PR #292, opened 2021-12-16
fix branch publishing

UnsungHero97:


geekbeast (approved):

geekbeast:
Admin override to fix publishing of artifacts.

PR #294, opened 2021-12-20
github actions

UnsungHero97:


geekbeast (dismissed):

UnsungHero97:
The build is too slow with the default GitHub Actions runner, 20+ minutes. We might revisit this in the future with a self-hosted runner but not worth spending time on this right now.

PR #302, opened 2022-01-12
Add additional artifactory repos

geekbeast:


geekbeast:
Admin override-- we need repos for successful builds since jcenter is down.

